PR #109, opened 2018-03-29
Allow mustache component to copy directories & add favicons to retail-app

ac360:
## What has been implemented?

The mustache template copying logic was broken, preventing you from being able to copy anything that was in a directory.

I also added favicons to the **retail-app** example.  These demonstrate that the mustache-app is working correctly.

## Steps to verify

* Run `components deploy` of the `retail-app` example included in this PR which includes the favicons.
* The favicons should be uploaded to S3 and you should see the favicon when viewing the retail-app in the browser.


registry/mustache/index.js
```diff
@@ -19,6 +19,9 @@ const deploy = async (inputs) => {
         path.relative(sourcePath, source.replace(/\.mustache$/i, ''))
       )
 
+      // Ensure tmp directories exist
+      fs.ensureDirSync(path.dirname(destination))
```
plfx: 💯

PR #110, opened 2018-03-29
Add env values to travis

brianneisler:
## What has been implemented?

Adds the necessary ENV values to the .travis.yml file for being able to publish to npm.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,9 @@ install:
 script:
 - npm run lint
 - npm test
+env:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -7,6 +7,9 @@ install:
 script:
 - npm run lint
 - npm test
+env:
```
ganeshvlrk: @brianneisler Are these considered to be sensitive. Do they need to be encrypted?

brianneisler: @ganeshvlrk i don't believe so since they get packaged up into our code anyways as part of the npm module

PR #114, opened 2018-03-30
Update GitHub components and add example

DavidWells:
## What has been implemented?

Closes https://serverlessteam.atlassian.net/browse/SC-112 & https://serverlessteam.atlassian.net/browse/SC-113

## Steps to verify

```
cd to serverless-components/examples/github-webhook-example

Add your config values

components deploy

Verify API gateway creation

Verify github webhook creation
```

## Todos:

* [ ] Write tests


DavidWells:
image

Value diffs

DavidWells:
image

.eslintrc.js
```diff
@@ -19,6 +19,10 @@ module.exports = {
     'comma-dangle': ['error', 'never'],
     'func-names': 'off',
     'no-param-reassign': 'off',
-    semi: ['error', 'never']
```
pmuens: IMHO we should stick to those rules rather than disabling them.

Prettier will auto-fix tons of linting issues (`prettier-atom` is the package I use. It auto-formats on save. Pretty slick...).

registry/github-webhook-aws/index.js
```diff
@@ -0,0 +1,13 @@
+
```
eahefnawy: why do we have those deploy and remove methods? are they required now?

eahefnawy: why do we have this entire file actually?

examples/github-webhook-example/code/handler.js
```diff
@@ -0,0 +1,11 @@
+
+module.exports.githubFunction = (event, context, callback) => {
+  console.log('Function ran!') // eslint-disable-line
+  return callback(null, {
+    statusCode: 201,
+    headers: {
+      'Content-Type': 'application/json'
+    },
+    body: 'created'
```
pmuens: Bikeshedding, but this is invalid JSON --> http://jsbin.com/sedinaxuni/edit?js,console

examples/github-webhook-example/serverless.yml
```diff
@@ -0,0 +1,17 @@
...
+    type: aws-lambda
+    inputs:
+      memory: 512
+      timeout: 10
+      handler: code/handler.githubFunction
+  GithubWebhookAWS:
+    type: github-webhook-aws
+    inputs:
+      function: ${webhookFunction}
+      githubRepo: DavidTron5000/other-repo
```
pmuens: Should we replace this with a more generic placeholder 🤔

examples/github-webhook-example/serverless.yml
```diff
@@ -0,0 +1,17 @@
...
+      githubApiToken: xx
```
pmuens: And this one too/

registry/github-webhook/index.js
```diff
@@ -0,0 +1,211 @@
+const octokit = require('@octokit/rest')()
+const R = require('ramda')
+const parseGithubUrl = require('parse-github-url')
+const diffValues = require('./diff')
+
+const createWebhook = async ({ githubApiToken, githubRepo, payloadUrl, events }) => {
+  octokit.authenticate({
+    type: 'token',
+    token: githubApiToken || process.env.GITHUB_TOKEN
```
eahefnawy: shouldn't we throw an early error if the token is not provided? why do have a placeholder?

package.json
```diff
@@ -31,6 +32,7 @@
     "axios": "^0.17.1",
     "bluebird": "^3.5.1",
     "chalk": "^2.3.2",
+    "concordance": "^3.0.0",
```
pmuens: Is diffing a new core-functionality? Otherwise this should be moved to the components dependencies.

registry/github-webhook-aws/index.js
pmuens: Unnecessary newline.

examples/github-webhook-example/code/handler.js
```diff
@@ -0,0 +1,11 @@
+
```
pmuens: Unnecessary newline.

registry/github-webhook-aws/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "@serverless/github-webhook-aws",
+  "version": "0.1.0",
+  "description": "Github webhook component using AWS API gateway",
+  "main": "index.js",
+  "author": "serverless.com",
+  "license": "MIT"
```
pmuens: This `package.json` doesn't include any dependencies. However the `package-lock.json` says that `parse-github-url` is a dependency. Either we should add this dependency here or remove `package.json`.

registry/github-webhook-aws/index.js
```diff
@@ -0,0 +1,13 @@
...
+async function deploy(inputs, context) {
+  return { ...inputs, ...context }
+}
+
+async function remove() {
```
pmuens: Do we need to add an "empty" `remove` function here? Or can we leave it and let the Framework handle the rest (since state are automatically wiped after removals)?

src/utils/variables/transformPostExecutionVars.js
```diff
@@ -15,6 +15,13 @@ module.exports = (slsYml) => {
         const referencedVariable = replace(/[${}]/g, '', reference).split('.')
         const referencedComponentAlias = referencedVariable[0]
         if (!reservedNames.includes(referencedComponentAlias)) {
+          if (!slsYml.components[referencedComponentAlias]) {
+            /* eslint-disable */
+            console.log(`Error: Unable to find "${referencedComponentAlias}" variable reference in ${slsYml.type} component.
+Please double check spelling of reference in ${slsYml.type} component.`)
+            /* eslint-enable */
+            // process.exit(1)
```
eahefnawy: why are we not throwing an error here? If we don't want to then this should be a warning imo. Also, we should remove that process.exit comment in that case.

src/utils/variables/transformPostExecutionVars.js
```diff
@@ -15,6 +15,13 @@ module.exports = (slsYml) => {
         const referencedVariable = replace(/[${}]/g, '', reference).split('.')
         const referencedComponentAlias = referencedVariable[0]
         if (!reservedNames.includes(referencedComponentAlias)) {
+          if (!slsYml.components[referencedComponentAlias]) {
+            /* eslint-disable */
+            console.log(`Error: Unable to find "${referencedComponentAlias}" variable reference in ${slsYml.type} component.
```
pmuens: Would `console.error` would be better here? 🤔

Not sure about CI / CD since it looks like this should proceed, if I'm reading the code here.

src/utils/variables/transformPostExecutionVars.js
pmuens: This should be removed.

src/utils/variables/transformPostExecutionVars.js
```diff
@@ -15,6 +15,13 @@ module.exports = (slsYml) => {
         const referencedVariable = replace(/[${}]/g, '', reference).split('.')
         const referencedComponentAlias = referencedVariable[0]
         if (!reservedNames.includes(referencedComponentAlias)) {
+          if (!slsYml.components[referencedComponentAlias]) {
+            /* eslint-disable */
```
pmuens: Is there any chance to only disable the specific rules here rather than disabling it globally and re-enabling it at the end?

Reason for this is that we might remove the `eslint-enable` comment below and have a bad time debugging this later on (already faced some similar issue in other Serverless projects).

scripts/doc-gen.js
```diff
@@ -0,0 +1,147 @@
+/* eslint-disable */
```
pmuens: ESLint should be enabled all the time. Prettier will help a lot with auto-fixing linting issues.

package.json
```diff
@@ -21,6 +21,7 @@
     "components": "./bin/components",
     "lint": "eslint . --cache",
     "test": "jest .",
+    "docs": "node scripts/doc-gen.js",
```
pmuens: This is golden and should auto-run on merges into `master` via Travis 👏 🎉

.eslintrc.js
DavidWells: This rule is not removed. Added more to make code easier to parse. 

Allow for explicit returns from arrow functions and `'object-shorthand': 'off'`

package.json
DavidWells: will be core I guess IDK up to product

registry/github-webhook-aws/index.js
DavidWells: IDK I thought we needed an index.js file.

DavidWells: Can we try and keep these code reviews about the code, this is way to tiny/insignificant to spend time reading and responding to

registry/github-webhook/index.js
DavidWells: Removed env defaults.

The component shouldn't run if required `githubApiToken` is missing.

IDK how this works with programatic access tho

scripts/doc-gen.js
DavidWells: Not here in a script that has nothing to do with core functionality

src/utils/variables/transformPostExecutionVars.js
DavidWells: linting fucks up the console output and a single line ignore would not work here

src/utils/variables/transformPostExecutionVars.js
DavidWells: done

PR #115, opened 2018-03-30
Add input to specify lambda package root

ac360:
## What has been implemented?

Fixes: https://serverlessteam.atlassian.net/browse/SC-193

This adds a `root` input to the `aws-lambda` component which allows the user to specify which directory will be zipped up and uploaded to AWS Lambda.

Before this, the `aws-lambda` component zipped up all files in your top-level component.  For example, in the `retail-app` component all files were being uploaded to AWS Lambda (e.g., the front-end assets).

This also updates the retail-app component's serverless.yml to NOT zip up everything.

## Steps to verify

* Deploy the retail app
* Go to the AWS Lambda console and see if your functions contain only the items in the `code` folder.


DavidWells:
I've reviewed the PR

I'm not sure we want to add another required property for this.

Instead we can probably infer the path based on the single property `handler` with path.dir or `path.resolve` and the lambda component where zip happens can under the hood infer the directory to zip up.

But not my call. Up to product.

ac360:
`root` is not required.

Inferring doesn't enable control of what to package.  Often, people deploy a couple nested folders, with the parent folder containing shared libs (e.g. package.json), across multiple functions.  Only inferring would result in loss of important functionality.

However, perhaps best of both worlds is to infer by default, if `root` is not included.  Right now it defaults to package up a `code` folder.

PR #119, opened 2018-03-30
Increase integration test timeouts

pmuens:
## What has been implemented?

Increase all the timeouts to avoid build failures.

## Steps to verify

Nothing to do here (other than looking at the code)...

## Todos:

* [x] Run Prettier


eahefnawy:
@pmuens why the tests are taking so long anyway? should we maybe mock something?

PR #120, opened 2018-03-30
Skip empty state removals

pmuens:
## What has been implemented?

Skips removals based on empty states which caused the framework to error-out in the past.

## Steps to verify

Deploy the `basic` example app. Remove it and run `remove` again.

## Todos:

* [x] Write tests
* [x] Run Prettier

### DIFF
--- a/src/utils/components/executeComponent.js
+++ b/src/utils/components/executeComponent.js
@@ -13,7 +13,6 @@ const executeComponent = async (
   rollback = false
 ) => {
   const component = components[componentId]
-  const context = generateContext(components, component, stateFile, archive, options, command)
   component.inputs = resolvePostExecutionVars(component.inputs, components)
 
   if (rollback) {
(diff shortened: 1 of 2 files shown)


src/utils/components/executeComponent.js
```diff
@@ -13,16 +13,19 @@ const executeComponent = async (
   rollback = false
 ) => {
   const component = components[componentId]
+  const context = generateContext(components, component, stateFile, archive, options, command)
```
brianneisler: @pmuens I think moving the call to generateContext here breaks the order of things. The generateContext call makes use of the component inputs. Does it not? By moving this here, the inputs are no longer properly resolved before being passed to generateContext.

pmuens: Dang. I think you're right.

Just wondering why our 3 integration tests pass 🤔... I'll look into that!

pmuens:

PR #121, opened 2018-03-30
Added S3 Tests (Includes a bug fix)

eahefnawy:
Includes a bug fix where no updates happen when the bucket name changes

### DIFF
--- a/registry/aws-iam-role/index.test.js
+++ b/registry/aws-iam-role/index.test.js
@@ -1,9 +1,38 @@
 const BbPromise = require('bluebird')
+const AWS = require('aws-sdk')
 const iamComponent = require('./index')
 
 jest.mock('bluebird', () => ({
   delay: jest.fn(() => Promise.resolve())
 }))
+
+jest.mock('aws-sdk', () => {
+  const mocks = {
+    createRoleMock: jest.fn().mockReturnValue({ Role: { Arn: 'abc:xyz' } }),
+    deleteRoleMock: jest.fn().mockReturnValue({ Role: { Arn: null } }),
+    attachRolePolicyMock: jest.fn(),
+    detachRolePolicyMock: jest.fn()
+  }
+
+  const IAM = {
+    createRole: (obj) => ({
+      promise: () => mocks.createRoleMock(obj)
+    }),
+    deleteRole: (obj) => ({
+      promise: () => mocks.deleteRoleMock(obj)
+    }),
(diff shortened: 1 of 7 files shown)


src/utils/components/generateContext.js
```diff
@@ -28,7 +28,7 @@ const generateContext = (
     serviceId,
     instanceId,
     provider: {
-      AWS
+      getSdk: () => AWS
```
plfx: I have deep reservations about this. We have another story to create an AWS provider component that outputs the configured SDK. It's actually implemented in https://github.com/serverless/serverless-components/tree/aws-component/registry/aws-provider, but questions came up about complications in usage that prevented it being merged.

Introducing the AWS SDK to the core is, in my opinion, a pretty strong stance to take just for the purpose of putting it in a mockable place for testing.

eahefnawy: @plfx Agreed. this is a temporarily solution proposed by @brianneisler until we get the provider components are in place.

registry/aws-s3-bucket/index.test.js
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    const inputs = {
+      name: 'some-bucket-name'
+    }
+
+    const outputs = await s3Component.deploy(inputs, s3ContextMock)
+
+    expect(AWS.S3).toHaveBeenCalledTimes(1)
+    expect(AWS.mocks.createBucketMock).toHaveBeenCalledTimes(1) // from the previous deploy
```
pmuens: Not how much effort it would be to implement this in different way, but inner-test-dependencies should be avoided as much as possible.

registry/aws-iam-role/index.test.js
```diff
@@ -21,7 +21,7 @@ describe('aws-iam-role unit tests', () => {
       log: () => {},
       saveState: () => {},
       provider: {
-        AWS: {
+        getSdk: () => ({
```
pmuens: I personally would vote for `getSDK` since abbreviations such as `JSON`, `AWS` or `API` are upper-cased in most cases. Not sure which naming pattern we follow here (if any),

registry/aws-s3-bucket/index.test.js
eahefnawy: yeah I don't think there's a way around it when it comes to mocking external modules. Since external modules are used in the entire file. Maybe we can make a require in every test? but that in itself is not great either.

PR #124, opened 2018-04-02
Add faker REST API example

rupakg:
## What has been implemented?

Add an example REST API to generate and return fake data.

Closes SC-174

## Steps to verify

* Run `deploy` to deploy resources
* See `README.md` for usage
* Run `remove` to remove deployed resources

## Todos:

* [ ] Write tests
* [ ] Run Prettier


.travis.yml
```diff
@@ -4,6 +4,7 @@ node_js:
 - 8
 install:
 - npm install
+- pushd examples/fakerapi/code && npm install && popd
```
plfx: This actually reveals a problem affecting our other examples; they're getting their modules from the core package, since we're not installing any.

We have a partial solution for this for registry modules, but it's not recursive so it still wouldn't catch this structure.  https://github.com/serverless/serverless-components/blob/master/scripts/postinstall.js#L17-L27

Rupak is to make a ticket to track us fixing that.

rupakg: Ticket created at: https://serverlessteam.atlassian.net/browse/SC-224

PR #125, opened 2018-04-02
Added Lambda Tests (Includes small bug fix)

eahefnawy:
Includes a small bug fix that handles removals when no deployment happened

### DIFF
--- a/registry/aws-lambda/index.test.js
+++ b/registry/aws-lambda/index.test.js
@@ -1,0 +1,212 @@
+const AWS = require('aws-sdk')
+const lambdaComponent = require('./index')
+
+jest.mock('./pack', () => jest.fn())
+
+jest.mock('aws-sdk', () => {
+  const mocks = {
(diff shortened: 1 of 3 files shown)


registry/aws-lambda/index.test.js
```diff
@@ -0,0 +1,252 @@
...
+
+jest.mock('aws-sdk', () => {
+  const mocks = {
+    createFunctionMock: jest.fn().mockReturnValue({ FunctionArn: 'abc:xyz' }),
+    updateFunctionConfigurationMock: jest.fn().mockReturnValue({ FunctionArn: 'abc:xyz' }),
+    updateFunctionCodeMock: jest.fn().mockReturnValue({ FunctionArn: 'abc:xyz' }),
+    deleteFunctionMock: jest.fn()
+  }
+
+  const Lambda = {
```
plfx: This set-up is heavy. These tests are still worthwhile, but we should find some way to make the mocks less piecemeal.

eahefnawy: Why is it heavy?

PR #127, opened 2018-04-03
Refactor pack tests

pmuens:
## What has been implemented?

Quick follow-up PR of https://github.com/serverless/serverless-components/pull/115 which ensures async-only code-usage, formatting according to prettier, doc updates and updated assertions.

## Steps to verify

Tests should still pass.

## Todos:

* [x] Write tests
* [x] Run Prettier

### DIFF
--- a/README.md
+++ b/README.md
@@ -696,6 +696,7 @@ Creates / Removes an AWS Lambda function.
 | `timeout`     | The functions timeout                | Number |
 | `description` | The functions description            | String |
 | `handler`     | The functions handler                | String |
+| `root`        | The code root directory              | String |
 | `role`        | The role `arn` the lambda should use | String |
 | `env`         | An object of env vars set at runtime | Object |
 
@@ -718,6 +719,7 @@ components:
       timeout: 10
       description: dome description
       handler: code.handler
+      root: code
       role:
         arn: arn:aws:iam::XXXXX:role/some-lambda-role
 ```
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -696,6 +696,7 @@ Creates / Removes an AWS Lambda function.
 | `timeout`     | The functions timeout                | Number |
 | `description` | The functions description            | String |
 | `handler`     | The functions handler                | String |
+| `root`        | The code root directory              | String |
```
DavidWells: I think we should remove this property and grab it automatically with `path.resolve` inside where the zip happens

plfx: I agree. We should just treat the directory containing the module as root. This keeps the config shorter:
```yml
handler: ${self.path}/code/functions.foo
# root is ${self.path}/code
# module is functions.js
# method is foo
```

Having them separately configurable (i.e. **not** allowing the module to be somewhere deep beneath the root, rather than in the root directly) allows weird usage that I've never seen before with JS. People will do things like this:
`root/utils.js`:
```js
// some shared file located in root?
module.exports = { stuff }
```
`root/handlers/foo.js`:
```js
const utils = require('../utils') // already weird
module.exports = {
  foo // and now we're exporting a handler from deep in the project, extra-weird
}
```

pmuens:
Merging this in since it's only a refactoring. We could open a separate issue for the suggested changes for the `root` property.

PR #130, opened 2018-04-03
Blog example app

rupakg:
## What has been implemented?
Implements a blog site that is generated via a static site generator Hugo.

Closes SC-175

## Steps to verify

* Deploy via `deploy`
* Remove via `remove`

## Todos:

* [ ] Write tests
* [ ] Run Prettier


registry/static-website/serverless.yml
```diff
@@ -10,6 +10,7 @@ inputTypes:
   templateValues:
     type: object
     default: {}
+    required: true  # problem with default value here SC-127
```
plfx: We were forced to flip this to "required" after Rupak discovered omitting this input does not work. Currently believe it is the same issue tracked here https://serverlessteam.atlassian.net/browse/SC-127

examples/blog/serverless.yml
```diff
@@ -0,0 +1,14 @@
+type: blog-app
+
+components:
+  blogSite:
+    type: static-website
+    inputs:
+      name: blog-site
+      contentPath: ${self.path}/site
+      templateValues: {}
```
plfx: This should not be required, but due to https://serverlessteam.atlassian.net/browse/SC-127 it is. It seems default values of type `object` simply don't apply correctly, and the framework sees unprovided object-type inputs as unresolved references, even with `required: false`.

PR #135, opened 2018-04-04
Manage own state on removal

pmuens:
## What has been implemented?

Prevent state-removals when errors occur during command execution.

## Steps to verify

Use the following service:

```yml
type: test-project

components:
  myTable:
    type: aws-dynamodb
    inputs:
      region: us-east-1
      tables:
        - name: my-table-${self.serviceId}
          hashKey: id
          indexes:
            - name: IdIndex
              type: global
              hashKey: id
          schema:
            id: number
            firstName: string
            lastName: string
          options:
            timestamps: true
```

Run `components deploy` (your table is not in a "create in progress" state) and after that immediately run `components remove`. You should see an error that the table could not be removed since it's in a pending state. The state shouldn't be wiped out if you look at the `state.json` file.

## Todos:

* [x] Write tests
* [x] Run Prettier

### DIFF
--- a/src/utils/components/executeComponent.js
+++ b/src/utils/components/executeComponent.js
@@ -39,7 +39,7 @@ const executeComponent = async (
 
   component.promise.resolve(component)
 
-  if (command === 'remove') {
+  if (command === 'remove' && !is(Function, func)) {
     stateFile[componentId] = {
       type: component.type,
       state: {}
--- a/src/utils/components/executeComponent.js
+++ b/src/utils/components/executeComponent.js
@@ -1,4 +1,6 @@
-const { is, isNil, isEmpty } = require('ramda')
+const {
+  not, is, isNil, isEmpty
+} = require('ramda')
 const generateContext = require('./generateContext')
 const resolvePostExecutionVars = require('../variables/resolvePostExecutionVars')
 const { getInputs, getState } = require('../state')
@@ -49,7 +51,7 @@ const executeComponent = async (
   kind of hackiness. There may be edge cases where this does not work, e.g.
   if the scripted component did not have any state keys to save in the first
   place, but for some reason still has complex removal logic. */
-  if (command === 'remove' && (!is(Function, func) || isNil(component.state))) {
+  if (command === 'remove' && (not(is(Function, func)) || isNil(component.state))) {
     stateFile[componentId] = {
       type: component.type,
       state: {}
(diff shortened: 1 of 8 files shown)


src/utils/components/executeComponent.test.js
```diff
@@ -136,22 +138,118 @@ describe('#executeComponent()', () => {
...
-      command,
-      options
-    )
-    expect(res.executed).toEqual(true)
-    expect(res.inputs).toEqual({
-      name: 'state-inputs-function-name',
-      memorySize: 128,
-      timeout: 5
+
+    it('should treat removals differently', async () => {
```
eahefnawy: I'd try to make this message more descriptive to what the test "should" behave 😊

pmuens:

I'll update the PR in a couple of minutes...

rupakg:
On @brianneisler point I thought we agreed that the component will throw a special kind of `exception` that is non-error and the core will just respond by NOT halting execution and NOT cleaning up state. At least that is what I thought was discussed.

src/utils/components/executeComponent.js
```diff
@@ -39,7 +41,17 @@ const executeComponent = async (
...
+  file if a remove was successful, if either the removed component has no
+  script, or else the script emptied the state during remove. This is required
+  to prevent certain features from breaking (e.g. skip removed components when
+  removing again) while also providing the ability for a component to
+  selectively report that its removal is incomplete pending manual and needs to
+  be run again. We need some deep architectural refactors to eliminate this
+  kind of hackiness. There may be edge cases where this does not work, e.g.
+  if the scripted component did not have any state keys to save in the first
+  place, but for some reason still has complex removal logic. */
+  if (command === 'remove' && (not(is(Function, func)) || isNil(component.state))) {
```
DavidWells: `not(is(Function, func))` is pretty gnarly to read.

vs vanilla JS

`typeof func !== 'function'`

IMO This is a good example of where not to use utils

Ramda shines in other areas though. Like deep equality checks and (others)

pmuens:
Thanks for the review and thanks for the feedback 👍 

🤔 I was not part of that conversation. Looking into the issue I can't find any note on the consensus we've agreed on. Maybe @plfx knows more here...

src/utils/components/executeComponent.js
plfx: I agree totally. I used it here because that's what this file uses for the same purpose above.

brianneisler:
Closing this as we went with the approach of throwing a special type of exception to maintain state

PR #137, opened 2018-04-05
WIP null check for promise

DavidWells:
## What has been implemented?

Closes https://serverlessteam.atlassian.net/browse/SC-203

## Steps to verify

https://serverlessteam.atlassian.net/browse/SC-203

## Todos:

* [ ] Write tests
* [ ] Run Prettier

### DIFF
--- a/src/utils/components/executeComponent.js
+++ b/src/utils/components/executeComponent.js
@@ -37,7 +37,9 @@ const executeComponent = async (
   }
   component.state = getState(stateFile, component.id)
 
-  component.promise.resolve(component)
+  if (component.promise) {
+    component.promise.resolve(component)
+  }
 
   if (command === 'remove') {
     stateFile[componentId] = {
--- a/src/utils/components/executeComponent.js
+++ b/src/utils/components/executeComponent.js
@@ -37,9 +37,7 @@ const executeComponent = async (
   }
   component.state = getState(stateFile, component.id)
 
-  if (component.promise) {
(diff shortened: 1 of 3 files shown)


DavidWells:
This fixes the error https://github.com/serverless/components/pull/137/files#diff-9a746a132c02bdc6dc50250494a89b66R40

However the deeper issue is that the shape of https://github.com/serverless/components/blob/edc8459ea15d9cbae19531e892c428ccca4a1083/src/utils/components/getComponentsToUse.js#L44-L52 is different than the shape of https://github.com/serverless/components/blob/master/src/utils/components/getComponentsToRemove.js#L31-L37

Anyone have insight into if this null check is good? Or do we need more here. @pmuens @eahefnawy @brianneisler

src/utils/components/executeComponent.js
```diff
@@ -37,7 +37,9 @@ const executeComponent = async (
   }
   component.state = getState(stateFile, component.id)
 
-  component.promise.resolve(component)
+  if (component.promise) {
```
plfx: I believe it should instead be safe to add a `promise` property to the component in `getComponentsToRemove.js`. It would look essentially the same as the one being added in `getComponentsToExecute`.

plfx:
Examples of needing such usage: the Dynamo component's removal might not complete. If the component was altered to have an output indicating whether its removal was completed, someone using that component might want to check and bail out on their own removal if it wasn't.

PR #138, opened 2018-04-05
Add signal event (SIGINT, SIGTERM, SIGBREAK) handling

pmuens:
## What has been implemented?

Implements the functionality to check again whether the user wants to force an exit when using `ctrl + c`.

Closes SC-208
Closes SC-223

## Acceptance criteria

- [x] Implement graceful shutdown --> Still interrupt process execution, but wrap up so that the state is still saved to disk

## Steps to verify

Deploy a service which might take some time to finish and hit `ctrl + c`. You should see a warning. Hitting `ctrl + c` again should force an exit.

## Todos:

* [x] Write tests
* [x] Run Prettier

### DIFF
--- a/src/utils/dag/executeGraph.js
+++ b/src/utils/dag/executeGraph.js
@@ -31,14 +31,11 @@ const execute = async (
       rollback
     )
     graph.removeNode(componentId)
-    return new Promise((resolve, reject) => {
-      const shouldGracefullyExit = getGracefulExitStatus()
-      if (shouldGracefullyExit) {
-        reject(new Error('Operation gracefully exited. State successfully persistet...'))
-      } else {
-        resolve(componentId)
-      }
-    })
+    const shouldGracefullyExit = getGracefulExitStatus()
+    if (shouldGracefullyExit) {
+      throw new Error('Operation gracefully exited. State successfully persistet...')
+    }
+    return componentId
   }, leaves)
 
   // allow all executions to complete without terminating
--- a/src/utils/dag/executeGraph.js
+++ b/src/utils/dag/executeGraph.js
@@ -1,8 +1,5 @@
 const { isEmpty, map } = require('ramda')
-const handleSignalEvents = require('../misc/handleSignalEvents')
 const executeComponent = require('../components/executeComponent')
-
-const { getGracefulExitStatus } = handleSignalEvents()
 
 const execute = async (
   graph,
@@ -31,9 +28,8 @@ const execute = async (
       rollback
     )
     graph.removeNode(componentId)
-    const shouldGracefullyExit = getGracefulExitStatus()
-    if (shouldGracefullyExit) {
-      throw new Error('Operation gracefully exited. State successfully persistet...')
+    if (global.signalEventHandling && global.signalEventHandling.shouldExitGracefully) {
+      throw new Error('Operation gracefully exited. State successfully persisted...')
     }
     return componentId
   }, leaves)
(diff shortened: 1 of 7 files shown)


src/utils/misc/handleSignalEvents.js
```diff
@@ -0,0 +1,30 @@
...
+  process.on('SIGINT', () => {
+    count += 1
+    const msg = [
+      'Waiting for current operations to finish... Press CTRL + C again to force an exit',
+      'NOTE: Doing so might corrupt the applications state information!'
+    ].join('\n')
+    if (count < 2) {
+      console.log(msg) // eslint-disable-line no-console
+    } else {
+      process.exit()
```
eahefnawy: should there be a specific exit code here?

pmuens: 🤔 good question. I'll dig deeper into that when implementing the graceful shutdown...

eahefnawy: +1

src/utils/dag/executeGraph.js
```diff
@@ -1,8 +1,18 @@
 const { isEmpty, map } = require('ramda')
+const handleSignalEvents = require('../misc/handleSignalEvents')
 const executeComponent = require('../components/executeComponent')
 
-const execute = async (graph, components, stateFile,
-  archive, command, options, rollback = false) => {
+const { getGracefulExitStatus } = handleSignalEvents()
```
brianneisler: @pmuens this kind of execution during import should really be avoided. This becomes very problematic during testing because there's no way to prevent the execution during import of ANY code where this file is a dependency.

src/utils/dag/executeGraph.js
```diff
@@ -21,6 +31,10 @@ const execute = async (graph, components, stateFile,
       rollback
     )
     graph.removeNode(componentId)
+    const shouldGracefullyExit = getGracefulExitStatus()
+    if (shouldGracefullyExit) {
+      throw new Error('Operation gracefully exited. State successfully persistet...')
```
brianneisler: minor misspelling of 'persisted'

src/utils/misc/handleSignalEvents.js
```diff
@@ -0,0 +1,48 @@
...
+    SIGINTCount += 1
+    const msg = [
+      'Waiting for current operation to gracefully finish (this might take some seconds)... Press CTRL + C again to force an exit',
+      'NOTE: Doing so might corrupt the applications state information!'
+    ].join('\n')
+    if (SIGINTCount < 2) {
+      console.log(msg) // eslint-disable-line no-console
+    } else {
+      gracefulExitStatus = false
+      process.exit()
```
brianneisler: I wonder if this should exit with non-zero?

src/utils/dag/executeGraph.js
pmuens: I see. Yes, that's also something I had problems with while writing the tests 😅

I'm doing this code-kung-fu here since I don't wanted to introduce a global variable and handle everything via Closures. I'll re-visit this decision and see what I can do here...

src/utils/dag/executeGraph.js
pmuens: Thanks 👍

src/utils/misc/handleSignalEvents.js
pmuens: Yes, that's a good idea. @eahefnawy pointed that out as well. We could go with `130` or `1`. I think I'll go with `130` 🤔

pmuens:
@brianneisler thanks for the review! I just updated the code accordingly 👍

pmuens:
So I implemented this via a `global` variable 👺

However the Promise is not rejected anymore if I interrupt the execution via `CTRL + C` 🤔.

pmuens:
D'oh. False alarm! I tested it again and it works 👍

I forgot to `await` the `BbPromise.delay()` in my mock function 🤦‍♂️

PR #140, opened 2018-04-06
Make IAM remove function idempotent

eahefnawy:
1. deploy an IAM component
2. copy the state
3. run remove the IAM component
5. run remove again.

### DIFF
--- a/registry/aws-iam-role/index.test.js
+++ b/registry/aws-iam-role/index.test.js
@@ -11,7 +11,7 @@ jest.mock('aws-sdk', () => {
     createRoleMock: jest.fn().mockReturnValue({ Role: { Arn: 'abc:xyz' } }),
     deleteRoleMock: jest.fn().mockImplementation((params) => {
       if (params.RoleName === 'some-already-removed-role') {
-        Promise.reject(new Error('Role not found'))
+        return Promise.reject(new Error('Role not found'))
       }
       return Promise.resolve({ Role: { Arn: null } })
     }),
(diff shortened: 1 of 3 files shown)


registry/aws-iam-role/index.test.js
```diff
@@ -9,7 +9,12 @@ jest.mock('bluebird', () => ({
 jest.mock('aws-sdk', () => {
   const mocks = {
     createRoleMock: jest.fn().mockReturnValue({ Role: { Arn: 'abc:xyz' } }),
-    deleteRoleMock: jest.fn().mockReturnValue({ Role: { Arn: null } }),
+    deleteRoleMock: jest.fn().mockImplementation((params) => {
+      if (params.RoleName === 'some-already-removed-role') {
+        Promise.reject(new Error('Role not found'))
```
pmuens: Shouldn't this `Promise` be returned here as well?

registry/aws-iam-role/index.test.js
```diff
@@ -135,4 +140,40 @@ describe('aws-iam-role unit tests', () => {
...
+      },
+      archive: {},
+      log: () => {},
+      saveState: jest.fn()
+    }
+
+    const outputs = await iamComponent.remove(inputs, iamContextMock)
+
+    expect(AWS.IAM).toHaveBeenCalledTimes(1)
+    expect(AWS.mocks.deleteRoleMock).toHaveBeenCalledTimes(2)
```
pmuens: Why is this called twice? Is it because of a dependency on a previous tests state?

PR #141, opened 2018-04-09
State file based removals

pmuens:
## What has been implemented?

Implements removals based on information stored in the state file rather than the `serverless.yml` file.

Closes SC-182

## Steps to verify

Deploy the following service (`myFunction` depends on `myRole`):

```yml
type: test-application

components:
  myFunction:
    type: tests-integration-function-mock
    inputs:
      name: my-function
      role: ${myRole.id}
  myRole:
    type: tests-integration-iam-mock
    inputs:
      name: execution-role
      service: my.function.service
```

Now change the `serverless.yml` so that it looks like this (now `myRole` depends on `myFunction`). **Don't deploy it**:

```yml
type: test-application

components:
  myFunction:
    type: tests-integration-function-mock
    inputs:
      name: my-function
  myRole:
    type: tests-integration-iam-mock
    inputs:
      name: ${myFunction.name}-execution-role
      service: my.function.service
```

Run `components remove`. You should be able to remove the service.

## Todos:

* [x] Write tests
* [x] Run Prettier


src/utils/components/getOrphanedComponents.js
```diff
@@ -0,0 +1,11 @@
+const { difference, keys, pick } = require('ramda')
+
+function getOrphanedComponents(serverlessFileComponents, stateFileComponents) {
+  const componentIdsInServerlessFile = keys(serverlessFileComponents)
+  const componentIdsInStateFile = keys(stateFileComponents)
+
+  const orphanedComponentIds = difference(componentIdsInStateFile, componentIdsInServerlessFile)
+  return pick(orphanedComponentIds, { ...serverlessFileComponents, ...stateFileComponents })
```
brianneisler: @pmuens i think orphaned components can only ever be from the state file. Don't think you need to merge the `serverlessFileComponents` and `stateFileComponents` here.

src/utils/components/getComponentsFromStateFile.js
```diff
@@ -0,0 +1,36 @@
+const { forEachObjIndexed, map, not, isEmpty, union, pickBy, prop } = require('ramda')
+const getComponentFunctions = require('./getComponentFunctions')
+const getChildrenIds = require('./getChildrenIds')
+const getState = require('../state/getState')
+const getInputs = require('../state/getInputs')
+const deferredPromise = require('../deferredPromise')
+
+async function getComponentsFromStateFile(stateFile) {
+  let componentIds = []
+  forEachObjIndexed((value, key) => {
```
brianneisler: @pmuens would be better to use `reduce` here and keep it immutable rather than mutating componentIds

src/utils/components/getComponentsFromStateFile.js
```diff
@@ -0,0 +1,36 @@
...
+    const state = getState(stateFile, key)
+    if (not(isEmpty(state)) && not(prop('internallyManaged', value))) {
+      componentIds = union(componentIds, [ key ])
```
brianneisler: You're only ever adding a single key to the `componentIds` here. It's much better to use `append` for that instead of joining using union. `union` tries to remove duplicates which is much slower and i don't think is ever a risk here?

src/utils/components/getComponentsFromStateFile.js
```diff
@@ -0,0 +1,36 @@
...
+    }
+  }, pickBy((k) => k !== '$', stateFile))
+  const componentsInfo = map(async (id) => {
```
brianneisler: Instead of using a `map` here, you could use `reduce` and get rid of the `Object.assign` from below. It's much cleaner and allows you to achieve the same result in a single pass instead of multiple.

src/utils/components/getComponentsFromServerlessFile.js
```diff
@@ -0,0 +1,61 @@
...
+  componentRoot = process.cwd(),
+  inputs = {},
+  componentId,
+  components = {}
+) => {
+  const component = await getComponent(componentRoot, componentId, inputs, stateFile)
+
+  const nestedComponents = await reduce(
+    async (accum, componentAlias) => {
+      accum = await Promise.resolve(accum)
```
brianneisler: This kind of promise chaining in a `reduce` is very dangerous. The reason is that it flattens out all calls into series instead of allowing them to run in parallel (since each iteration of the reduction has to wait for the accumulator to resolve before it can proceed).

src/utils/components/getOrphanedComponents.js
pmuens: Hehe. I thought the same but then I thought "Better safe than sorry". Will update it 👍

src/utils/components/getComponentsFromStateFile.js
pmuens: Good point. Updating...

src/utils/components/getComponentsFromStateFile.js
pmuens: Alright 👍 I avoided reducing since this operation is done on an object, but I figured another way to `reduce` here.

src/utils/components/getComponentsFromStateFile.js
pmuens: Good point! +1 for that 👍

src/utils/components/getComponentsFromServerlessFile.js
pmuens: 🤔 I think I understand the root issue here. How can we fix this? (Aside: I'm not the author of that code...)

brianneisler:
@pmuens changes look good but requires merging resolving conflicts 🤦‍♂️

pmuens:
Thanks for approving @brianneisler 👍 I'll rebase (and hopefully merge) today!

PR #143, opened 2018-04-10
File system referencing component types

pmuens:
## What has been implemented?

Implements the functionality so that users can reference local components with the help of file system directives in the components `type` property.

Closes SC-59

---

**NOTE:** TBH I'm not sure if sotring an abosulte path in the `state.json` file is the best approach here. Main concern is that the state file is unusable if stuff is moved around. Not sure if that's a show-stopper / how other tools solve such issues 🤔

---

**Todos:**

- [x] Store relative paths in the `state.json` file (rather than an absolute path)

## Steps to verify

Create the following directory with the given files somewhere on your local machine (the only place where you shouldn't store it is in the `registry` directory).

```yml
# custom-components/my-function/serverless.yml

type: my-function

inputTypes:
  name:
    type: string
    required: false
    default: my-function-${self.instanceId}
```

```js
// custom-components/my-function/index.js

function deploy(inputs, context) {
  context.log(`Deploying custom function "${inputs.name}"...`)
  const newState = {
    ...context.state,
    updatedAt: +new Date()
  }
  context.saveState(newState)

  return newState
}

function remove(inputs, context) {
  context.log(`Removing custom function "${inputs.name}"...`)
  context.saveState()
}

module.exports = {
  deploy,
  remove
}
```

Now create the following test service:

```yml
type: test-application

components:
  myCustomFunction:
    # NOTE: change the following to the path of your custom component
    type: ../../custom-components/my-function
    inputs:
      name: custom-function-name-${self.instanceId}
  myTestFunction:
    type: tests-integration-function-mock
    inputs:
      name: test-function-name-${self.instanceId}
```

Deploy it. You should see `console.log` statements from both functions and the corresponding state information in the `state.json` file.

## Todos:

* [x] Write unit tests
* [x] Run Prettier


src/utils/components/getOrphanedComponents.js
```diff
@@ -5,7 +5,7 @@ function getOrphanedComponents(serverlessFileComponents, stateFileComponents) {
   const componentIdsInStateFile = keys(stateFileComponents)
 
   const orphanedComponentIds = difference(componentIdsInStateFile, componentIdsInServerlessFile)
-  return pick(orphanedComponentIds, stateFileComponents)
+  return pick(orphanedComponentIds, { ...serverlessFileComponents, ...stateFileComponents })
```
pmuens: This should be changed back to only pick from the `stateFileComponents`.

eahefnawy:
hmmm on second thought, once this local component is published to our future registry, this type property would be required.

pmuens:
Just updated the PR so that relative paths are stored in the `state.json` file (based on our discussion yesterday). Nothing else was changed...

This PR is ready for another (final) review.

@eahefnawy it would be super nice if you can take a second look and merge if you think it's good to go. Thanks 👍

PR #152, opened 2018-04-17
Make cloudfront component removal idempotent

eahefnawy:
1. deploy the component
2. copy the state
3. run remove the component
5. run remove again.

### DIFF
--- a/registry/aws-cloudfront/index.js
+++ b/registry/aws-cloudfront/index.js
@@ -203,18 +203,12 @@ const deleteDistribution = async (distributionId) => {
   // disable distribution first
   const res = await toggleEnabledForDistribution(distributionId, false)
 
-  // delete distribution
-  try {
-    await CloudFront.deleteDistribution({
-      Id: distributionId,
-      IfMatch: res.distribution.eTag
-    }).promise()
-    console.log(`CloudFront distribution '${distributionId}' deletion initiated`)
-    return {}
-  } catch (err) {
-    console.log(`Error in deleting CloudFront distribution '${distributionId}'`, err.message)
-    return { error: err }
-  }
+  await CloudFront.deleteDistribution({
+    Id: distributionId,
+    IfMatch: res.distribution.eTag
+  }).promise()
+
+  console.log(`CloudFront distribution '${distributionId}' deletion initiated`)
 }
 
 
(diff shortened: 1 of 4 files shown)


registry/aws-cloudfront/index.js
```diff
@@ -248,14 +242,19 @@ const deploy = async (inputs, context) => {
...
   if (!context.state.name) return {}
 
-  context.log(`Removing CloudFront distribution: '${context.state.name}' with id: '${context.state.distribution.id}'`)
-  const res = await deleteDistribution(
-    context.state.distribution.id,
-    context.state.distribution.eTag
-  )
-  if (!res.error) {
-    context.saveState({})
+  try {
```
brianneisler: @eahefnawy why include this logic here instead of in the helper method? Is there ever a case where we would NOT want to treat the deletion in an idempotent way?

registry/aws-cloudfront/index.js
```diff
@@ -248,14 +242,19 @@ const deploy = async (inputs, context) => {
...
+    context.log(`Removing CloudFront distribution: '${context.state.name}' with id: '${context.state.distribution.id}'`)
+    await deleteDistribution(
+      context.state.distribution.id,
+      context.state.distribution.eTag
+    )
+  } catch (e) {
+    if (!e.message.includes('The specified distribution does not exist')) {
+      throw new Error(e)
```
brianneisler: `throw e` instead

registry/aws-cloudfront/index.js
```diff
@@ -248,14 +242,19 @@ const deploy = async (inputs, context) => {
...
+    }
   }
+
+  context.saveState({})
```
brianneisler: This is the default, no need to do this here.

registry/aws-cloudfront/index.js
eahefnawy: Because the error actually comes from the `CloudFront.getDistribution` aws call, which is in the `getDistribution` helper method (not the `deleteDistribution` helper method as we've been handling it before), which is called in two other helper methods. Putting it at the highest level at the `remove` function makes sure we catch it regardless of where it originates from.

eahefnawy:
@brianneisler I've pushed some updates based on your comments and added some tests to the core and the cloudfront component

PR #153, opened 2018-04-17
Make Route53 component remove function idempotent

eahefnawy:
1. deploy component
2. copy the state
3. run remove the component
5. run remove again.


eahefnawy:
@brianneisler I've pushed some updates based on your comments and added tests

PR #160, opened 2018-04-19
Add Codecov support

pmuens:
## What has been implemented?

Adds Codecov support so that we can see the actual code coverage and add guards to prevent untested code from being merged.

Closes SC-236

---

I tried to add the GitHub integration, but it looks like I'm not authorized to do that. @brianneisler it would be nice if you could add that to the repo so that we can have PR / Branch checks.

## Todos:

* [x] Run Prettier

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,23 @@
+codecov:
+  notify:
+    require_ci_to_pass: yes
+
+coverage:
+  precision: 2
+  round: down
+  range: "70...100"
+
+  status:
+    project: yes
+    patch: yes
+    changes: no
+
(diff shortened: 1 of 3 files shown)


eahefnawy:
What's our acceptance threshold when it comes to code coverage? I think at this point a big portion of the codebase is still not covered, no?

pmuens:
@eahefnawy good question 🤔 I just added a `.codecov.yml` which disables the annoying PR comments and adds a threshold of at least 70 percent (which is also the default). This should work as a baseline. We could tweak it if we have more tests and the coverage is increased.

PR #161, opened 2018-04-19
Auto-run ESLint and Prettier on every Git commit

pmuens:
## What has been implemented?

Automatically runs ESLint and Prettier on every Git commit, ensuring that only linted and prettified files are pushed.

Closes SC-244
Closes SC-245

## Steps to verify

Check out the branch. Create a new branch based on this branch. Modify a .js file (add ESLint violations and format it slightly). Stage the file and commit it. The pre commit checks should kick in --> ESLint and Prettier should auto-fix violations when possible.

## Todos:

* [x] Run Prettier

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,3 +7,4 @@ services:
     volumes:
       - .:/app
       - ~/.aws/:/root/.aws
+      - ~/.gitconfig:/root/.gitconfig
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,45 +5,58 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.0.0-beta.46",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0-beta.46.tgz",
-      "integrity": "sha512-7BKRkmYaPZm3Yff5HGZJKCz7RqZ5jUjknsXT6Gz5YKG23J3uq9hAj0epncCB0rlqmnZ8Q+UUpQB2tCR5mT37vw==",
+      "version": "7.0.0-beta.40",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0-beta.40.tgz",
+      "integrity": "sha512-eVXQSbu/RimU6OKcK2/gDJVTFcxXJI4sHbIqw2mhwMZeQ2as/8AhS9DGkEDoHMBBNJZ5B0US63lF56x+KDcxiA==",
       "dev": true,
       "requires": {
-        "@babel/highlight": "7.0.0-beta.46"
+        "@babel/highlight": "7.0.0-beta.40"
       }
     },
     "@babel/highlight": {
-      "version": "7.0.0-beta.46",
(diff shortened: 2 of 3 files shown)


eahefnawy:
let's not do that 😞 ... sometimes we need to make quick pushes and that could be a blocker. There are cases where eslint will succeed on Travis but fail locally (ie. if I didn't run `npm install` locally for a quick doc change).

I think we should strike a balance when it comes to code quality. Have Github checks imo is enough to make sure **master** is at high quality, not the branches.

brianneisler:
@eahefnawy when do we need to push unlinted code? Having to fix someone else's code because they forgot to address these problems in their own branch is a real pain for other developers.

eahefnawy:
@brianneisler Here's the situation that I keep running into, specially in the components repo (I've run into the same with the SDK & Platform repos since they also have many node_modules folder)

- I make changes in the core, and make sure all relavent linting issues are fixed.
- However, when I run eslint locally, I often get lint errors from other component directories (which include their own node_modules folder) because I didn't run npm install in each of them because someone else in the team updated the dependencies. The error I usually get is "cannot find module x" in those components directory.
- Because I know why this is happening, I push my code to Travis anyway, and the build of course succeed because travis install all dependencies.

Of course solving that is as simple as keep running npm i in each component to make sure **all** dependencies are up to date. However imo that causes unnecessary friction when it **must** be done on each commit, specially that I know the build will succeed. This is especially true when you make quick docs changes.

We can add an automated script for that of course, but again, it's still a friction having to wait for all npm modules in all components to install.

Imo just using Travis for lint validation should strike a good balance between code quality and efficiency.

pmuens:
Thanks for the feedback @eahefnawy and @brianneisler 👍

I understand that this setup can slow one down when quick fixes should be applied, but I think that this is a good improvement considering the long-term value we get out of it.

My feeling is that I'm one of very few peoples actually using Prettier and ESLint locally.
I've setup and integrated Prettier and ESLint with some Atom plugins and the workflow is smooth and unbotrusive.

Way too often I've seen `fix lint`, `fixing lint`, `yet another lint fix`, commits which are unnecessary since PRs where linting violations are present can't be merged and adding dozens of commits to fix trivial linting issues pollutes the git history.

Other than that I faced re-formatting of already commited files due to non-prettier usage countless of times. This results in bloated PRs which are hard to review since the formatting was also changed while functionality was added / removed.

IMHO it sounds more like a local dev setup issue rather than a show-stopper to merge this PR. Maybe there are some ways (e.g. scripts) to streamline the setup and help people integrate this in their editors.

pmuens:
@eahefnawy @brianneisler any decision regarding this PR? Should we merge or close it

(I'd rebase it if we want to merge it)

pmuens:
> @pmuens ok, could you add an npm script that installs all component dependencies? this should reduce the friction a little bit and make the experience better 😊

@eahefnawy the `postinstall` script should do that already 🤔. So `npm run postinstall` should install all the component dependencies (for the components in the registry).

pmuens:
@eahefnawy @brianneisler Just rebased the PR and resolved the conflicts.

eahefnawy:
@pmuens that works. Merging... 😊

PR #162, opened 2018-04-19
Added unit tests for APIG component remove function

eahefnawy:


registry/aws-apigateway/index.test.js
```diff
@@ -0,0 +1,90 @@
...
+        id: 'throwError'
+      },
+      archive: {},
+      log: () => {},
+      saveState: () => {}
+    }
+
+    await expect(apigComponent.remove({}, apigContextMock)).rejects.toThrow('some random aws error')
+    expect(AWS.APIGateway).toHaveBeenCalledTimes(1)
+    expect(AWS.mocks.deleteRestApiMock).toHaveBeenCalledTimes(3)
```
brianneisler: @eahefnawy it looks like the number of expected times to be called is going up per test because the tests are sharing the same counters. It's good practice to ensure that tests in NO WAY impact each other's assertions. To fix this you can simply use the `mockClear` method to reset the counters before each test https://facebook.github.io/jest/docs/en/mock-function-api.html#mockfnmockclear

pmuens: This might be handy as well (since it can be used to clear all mocks at once): https://facebook.github.io/jest/docs/en/jest-object.html#jestclearallmocks

eahefnawy: Thank you both for the feedback! 😊 

I've used mockClear in an afterEach block. Works magical!

@pmuens I tried to use clearAllMocks in an after each but it didnt work. I think because the aws-sdk manual mock gets clear as well. I tried to put it in a beforeEach but that didn't work either.

eahefnawy:
Thanks guys. I've pushed the mockClear change, and I've also opened a new PR that implements this change for the other aws components tests as well:

https://github.com/serverless/components/pull/169

PR #164, opened 2018-04-20
Added package command and core version validation

eahefnawy:
From within the basic example directory, run the following:

```
components package --path ./
```

```
components package --path ./ --format tar
```

### DIFF
--- a/examples/basic/serverless.yml
+++ b/examples/basic/serverless.yml
@@ -1,4 +1,5 @@
 type: my-project
+version: 0.0.1
 
 components:
   myFunction:
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "raven": "^2.4.2",
     "rc": "^1.2.6",
     "s3-client": "^4.4.1",
-    "semver-regex": "^2.0.0",
+    "semver": "^5.5.0",
     "uuid": "^3.2.1",
     "write-file-atomic": "^2.3.0"
   },
--- a/package.json
+++ b/package.json
@@ -29,7 +29,6 @@
   },
   "dependencies": {
     "@octokit/rest": "^14.0.5",
-    "adm-zip": "^0.4.8",
     "analytics-node": "^3.2.0",
     "archiver": "^2.1.1",
     "aws-sdk": "^2.214.1",
@@ -56,6 +55,7 @@
     "write-file-atomic": "^2.3.0"
   },
   "devDependencies": {
+    "adm-zip": "^0.4.9",
     "babel-jest": "^22.4.1",
     "eslint": "^4.18.1",
     "eslint-config-airbnb-base": "^12.1.0",
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
   "dependencies": {
     "@octokit/rest": "^14.0.5",
     "analytics-node": "^3.2.0",
+    "archiver": "^2.1.1",
     "aws-sdk": "^2.214.1",
     "axios": "^0.17.1",
     "bluebird": "^3.5.1",
@@ -49,6 +50,7 @@
     "raven": "^2.4.2",
     "rc": "^1.2.6",
     "s3-client": "^4.4.1",
+    "semver-regex": "^2.0.0",
     "uuid": "^3.2.1",
     "write-file-atomic": "^2.3.0"
   },
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
   },
   "dependencies": {
     "@octokit/rest": "^14.0.5",
+    "adm-zip": "^0.4.8",
     "analytics-node": "^3.2.0",
     "archiver": "^2.1.1",
     "aws-sdk": "^2.214.1",
--- a/src/utils/components/getComponent.js
+++ b/src/utils/components/getComponent.js
@@ -1,6 +1,8 @@
 const path = require('path')
 const { forEachObjIndexed } = require('ramda')
 const { readFile } = require('../fs')
+const validateCoreVersion = require('../validateCoreVersion')
+
 const getServiceId = require('../state/getServiceId')
 const transformPostExecutionVars = require('../variables/transformPostExecutionVars')
 const resolvePreExecutionVars = require('../variables/resolvePreExecutionVars')
@@ -13,6 +15,7 @@ module.exports = async (componentRoot, componentId, inputs, stateFile) => {
   let slsYml = await readFile(path.join(componentRoot, 'serverless.yml'))
 
   validateVarsUsage(slsYml)
+  validateCoreVersion(slsYml.type, slsYml.core)
 
   slsYml.id = componentId || slsYml.type
 
--- a/src/utils/components/packageComponent.js
+++ b/src/utils/components/packageComponent.js
@@ -1,0 +1,34 @@
+const path = require('path')
+const semverRegex = require('semver-regex')
+
+const { fileExists, readFile } = require('../fs')
+const pack = require('../pack')
+
+module.exports = async (options) => {
+  const validFormats = [ 'zip', 'tar' ]
+  const format = options.format || 'zip'
+  const slsYmlFilePath = path.join(process.cwd(), 'serverless.yml')
+  if (!await fileExists(slsYmlFilePath)) {
+    throw new Error('The package command can only be run inside a component directory')
+  }
+
+  const slsYml = await readFile(slsYmlFilePath)
+
+  if (!semverRegex().test(slsYml.version)) {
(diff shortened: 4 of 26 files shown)


eahefnawy:
I'll add tests once an initial review has been made

eahefnawy:
Note: This is the first core command and there's currently no structured way to define core commands. Once we've implemented @DavidWells proposal on defining CLI commands (both core and components commands), this should be updated.

src/utils/components/packageComponent.js
```diff
@@ -0,0 +1,34 @@
...
+
+  if (!validFormats.includes(format)) {
+    throw new Error('Please provide a valid format. Either a "zip" or a "tar"')
+  }
+
+  const outputFileName = `${slsYml.type}@${slsYml.version}.${format}`
+  const outputFilePath = path.resolve(options.path, outputFileName)
+
+  return pack(process.cwd(), outputFilePath, format)
+    .then(() => console.log(`Component has been packaged in ${outputFilePath}`)) // eslint-disable-line
```
pmuens: We should re-use the `log` method we also attach to the `context` here.

src/utils/components/packageComponent.js
```diff
@@ -0,0 +1,34 @@
+const path = require('path')
+const semverRegex = require('semver-regex')
+
+const { fileExists, readFile } = require('../fs')
+const pack = require('../pack')
+
+module.exports = async (options) => {
+  const validFormats = [ 'zip', 'tar' ]
+  const format = options.format || 'zip'
+  const slsYmlFilePath = path.join(process.cwd(), 'serverless.yml')
```
pmuens: Would be nice to refactor this into a separate method (including the check below) since we might want to re-use this in a couple of other places too.

src/utils/components/packageComponent.js
```diff
@@ -0,0 +1,34 @@
...
+    throw new Error('The package command can only be run inside a component directory')
+  }
+
+  const slsYml = await readFile(slsYmlFilePath)
+
+  if (!semverRegex().test(slsYml.version)) {
+    throw new Error('Please provide a valid version for your component')
+  }
+
+  if (!options.path) {
```
pmuens: Should we support `p` as well?

src/utils/components/packageComponent.js
```diff
@@ -0,0 +1,34 @@
...
+    throw new Error('Please provide an output path for the package with the --path option')
+  }
+
+  if (!validFormats.includes(format)) {
+    throw new Error('Please provide a valid format. Either a "zip" or a "tar"')
```
pmuens: Quick feedback: Naturally I would've entered `.zip` rather than `zip` or `.tar` rather than `tar`.

examples/basic/serverless.yml
```diff
@@ -1,4 +1,5 @@
 type: my-project
+version: 0.0.1
```
pmuens: Where is this property used?

src/utils/components/packageComponent.js
eahefnawy: hmm I thought this function is meant to be used exclusively by components. As long as there are no situations where we want to control/disable component logs but not core logs, then you are right.

I'm gonna make that change though as it's easy to revert

src/utils/components/packageComponent.js
eahefnawy: I think we should think about shortcuts once we nail down the details of the new CLI implementation we're gonna create.

src/utils/components/packageComponent.js
eahefnawy: hmmm disagreed. This option is called `format` not `extension`. Also, an extra dot is a little more friction.

src/utils/components/packageComponent.js
eahefnawy: 👍

examples/basic/serverless.yml
eahefnawy: during validation here:
https://github.com/serverless/components/blob/package-command/src/utils/components/packageComponent.js#L17

eahefnawy: and a little down under too to generate the package file name

src/utils/components/packageComponent.js
pmuens: Yes, I thought about the same - Maybe we need another `log` method for core functionalities. We could discuss that later on when we see a need for it 🤔

src/utils/components/packageComponent.js
pmuens: Yes, good idea 👍

examples/basic/serverless.yml
pmuens: I see. Wouldn't it be necessary in the other examples as well?

eahefnawy: Yep, you're right. Actually this should be updated in all components in the registry too.

Okay, how about this: I'll update the examples in this PR, and will update the registry as part of the work on SC-251, which will also include more complex changes to support versions in `type` as well when you're using nested components.

eahefnawy: very good catch btw 👍 ... not updating the examples would break all examples after merging.

src/utils/components/packageComponent.js
eahefnawy: hmmmm on second thought while adding this, wrapping a single line into its own function and file is an overhead that doesn't make sense. This is especially true for the check below as that would just turn `await fileExists(slsYmlFilePath)` into `await slsYmlExists(slsYmlFilePath)` plus an extra file that is also required in the util root.

examples/basic/serverless.yml
pmuens: That sounds reasonable :+1:

src/utils/pack.test.js
```diff
@@ -0,0 +1,44 @@
...
+    await fsp.ensureDirAsync(outputDirPath)
+    await fsp.writeJsonAsync(path.join(inputDirPath, 'foo.json'), {
+      key1: 'value1',
+      key2: 'value2'
+    })
+  })
+
+  it('should package a directory and return the file path', async () => {
+    const outputFilePath = path.join(outputDirPath, `${crypto.randomBytes(6).toString('hex')}.zip`)
+    const returnedoutputFilePath = await pack(inputDirPath, outputFilePath)
```
pmuens: `returnedOutputFilePath`

src/utils/pack.test.js
```diff
@@ -0,0 +1,44 @@
...
+      content: entry.getData()
+    }))
+    const jsonFile = files.filter((file) => file.name === 'foo.json').pop()
+
+    expect(files.length).toEqual(1)
+    expect(JSON.parse(jsonFile.content.toString('utf8'))).toEqual({
+      key1: 'value1',
+      key2: 'value2'
+    })
+    expect(returnedoutputFilePath).toEqual(outputFilePath)
```
pmuens: `returnedOutputFilePath`

package.json
```diff
@@ -28,7 +28,9 @@
   },
   "dependencies": {
     "@octokit/rest": "^14.0.5",
+    "adm-zip": "^0.4.8",
```
pmuens: This should be a `devDependency` since we're only using it in our tests.

src/utils/components/getComponent.js
```diff
@@ -11,6 +13,8 @@ const validateInputs = require('./validateInputs')
 module.exports = async (componentRoot, componentId, inputs, stateFile) => {
   let slsYml = await readFile(path.join(componentRoot, 'serverless.yml'))
 
+  validateCoreVersion(slsYml.type, slsYml.core)
```
pmuens: Why isn't the second parameter `slsYml.version`? Where is the `slsYml.core` coming from? 🤔

src/utils/pack.test.js
eahefnawy: NICE CATCH!!!! ♥️

package.json
eahefnawy: yep! spot on! 💯

src/utils/components/getComponent.js
eahefnawy: This PR introduces two new properties to serverless.yml:

- **version:** this is the version of the component itself. It is only validated on packaging to make sure that it's a valid semver that can be used later in publishing and also to make sure the package file name is consistent.
- **core:** this is the core version this component is compatible with. We do this validation in the line above.

pmuens: Ah. I see! Thanks for the clarification. Shouldn't we add the `core` value to the examples as well?

eahefnawy: nope, `core` is optional

PR #166, opened 2018-04-23
Fix AppVeyor setup

pmuens:
## What has been implemented?

Fixes the issues we're currently having with AppVeyor.

~**NOTE:** I updated everything so that it runs on Windows. It looks like Appveyor is super slow ATM causing our integration tests to timeout. I raised the timeout-limit to 60 seconds w/o success. Maybe we should disable integration tests on Appveyor 🤔~

## Steps to verify

Take a look into the AppVeyor build logs.

## Todos:

* [x] Run Prettier

### DIFF
--- a/tests/integration/await-child-components.test.js
+++ b/tests/integration/await-child-components.test.js
@@ -44,7 +44,7 @@ describe('Integration Test - Await child components', () => {
 
   describe('when running through a typical component usage lifecycle', () => {
     it('should deploy the "function" component and await its deployment', async () => {
-      await cpp.execAsync(`${componentsExec} deploy`, {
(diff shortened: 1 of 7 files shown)


tests/integration/await-child-components.test.js
```diff
@@ -44,7 +44,7 @@ describe('Integration Test - Await child components', () => {
 
   describe('when running through a typical component usage lifecycle', () => {
     it('should deploy the "function" component and await its deployment', async () => {
-      await cpp.execAsync(`${componentsExec} deploy`, {
+      await cpp.execAsync(`node ${componentsExec} deploy`, {
```
eahefnawy: ha! nice catch!! 😀

PR #167, opened 2018-04-24
Simplify Component API

DavidWells:
## What has been implemented?

```js
// example of new component index.js
export const Create = (inputs, context) => {}
export const Update = (inputs, context) => {}
export const Delete = (inputs, context) => {}
```

Closes https://serverlessteam.atlassian.net/browse/SC-254

## Steps to verify

```
cd examples/github-webhook-example
../../bin/components deploy
# Change repo url
# then redeploy
../../bin/components deploy
# resource cleans itself via `defaultDeploy` function
```

Default deploy method if one is not exported from component https://github.com/serverless/components/pull/167/files#diff-9a746a132c02bdc6dc50250494a89b66R88


registry/github-webhook/github.js
```diff
@@ -0,0 +1,135 @@
...
+    // already exists. Handle idempotence
+    if (e.code === 422) {
+      // Get hooks.
+      const hooks = await octokit.repos.getHooks({
+        owner: gh.owner,
+        repo: gh.name,
+        page: 1 // Must be updated to handle recursive lookups
+      })
+
+      if (hooks.data) {
```
pmuens: You can use Ramdas filter capabilites to include this check for presence of `hooks.data` in the `filter` call since Ramda won't throw if there's an empty array:

http://jsbin.com/cemofiqupu/edit?js,console

registry/github-webhook/github.js
```diff
@@ -0,0 +1,135 @@
...
+          return urlMatch && eventsMatch
+        })
+        if (match && match.length) {
+          hookData = match[0]
+        }
+      }
+    }
+
+    if (e.code === 404) {
+      /*eslint-disable */
```
pmuens: I'd suggest to disable liniting on a per-line basis given the fact that in this case only one line is disabled.

The reason for that is that if the code within this blog expands and we move things around we might accidentally remove the `eslint-enable` comment causing errors to slip thourgh the cracks.

registry/github-webhook/github.js
```diff
@@ -0,0 +1,135 @@
...
+    active: true
+  }
+  let hookData
+  try {
+    const res = await octokit.repos.editHook(params)
+    hookData = res.data
+  } catch (e) {
+    // If API errors
+    if (e.code === 404) {
+      /* eslint-disable */
```
pmuens: Same here.

registry/github-webhook/github.js
```diff
@@ -0,0 +1,135 @@
...
+  try {
+    await octokit.repos.deleteHook({
+      owner: gh.owner,
+      repo: gh.name,
+      id: id
+    })
+  } catch (e) {
+    // If API errors
+    if (e.code === 404) {
+      /* eslint-disable */
```
pmuens: Same here.

registry/github-webhook/index.js
```diff
@@ -170,42 +50,8 @@ function webhookExists(context) {
...
-      }
-    }
-    return acc
-  }, initialData)
-}
-
 module.exports = {
-  deploy,
-  remove
+  Create,
```
pmuens: I'd suggest to stick to the `lowerCase` convention we're using throughout the codebase (if there's not reason to use `UpperCase`).

src/utils/components/compareInputsToState.js
```diff
@@ -0,0 +1,49 @@
+const R = require('ramda')
```
pmuens: It'd be better to only import the necessary parts or Ramda so that we can tree-shake later on.

src/utils/components/compareInputsToState.js
```diff
@@ -0,0 +1,49 @@
...
+    // If no state keys... no state
+    hasState,
+    // default everything is equal. Zen
+    isEqual: true,
+    // keys of state that are different
+    keys: [],
+    // if differences, diffs values of inputs vs state
+    diffs: {}
+  }
+  return Object.keys(inputs).reduce((acc, current) => {
```
pmuens: Since Ramda is used anyway we could use Ramdas reduce here.

src/utils/components/executeComponent.js
```diff
@@ -26,15 +28,23 @@ const executeComponent = async (
 
   if (command === 'remove') {
     if (isEmpty(context.state)) {
+      console.log('empty?')
```
pmuens: This should be removed.

src/utils/components/executeComponent.js
```diff
@@ -44,6 +54,25 @@ const executeComponent = async (
     }
     component.executed = true
   }
+
+  //*
```
pmuens: Should this be included as commented-out or commented code?

src/utils/components/executeComponent.js
```diff
@@ -44,6 +54,25 @@ const executeComponent = async (
...
+  // turn off random exported commands
+    const func = component.fns[command]
+    //let retainState = false
```
pmuens: Do we use this? Otherwise we might want to remove it.

src/utils/components/executeComponent.js
```diff
@@ -58,4 +87,98 @@ const executeComponent = async (
   return component
 }
 
+async function defaultDeploy(inputs, context, component) {
+  const fns = component.fns
+  const inputTypes = component.inputTypes
+
+  const componentData = compareInputsToState(inputs, context.state)
+  // console.log('component diff data', componentData)
```
pmuens: Looks like we can remove this.

src/utils/components/executeComponent.js
```diff
@@ -58,4 +87,98 @@ const executeComponent = async (
...
+  const inputsChanged = !componentData.isEqual
+  // console.log('inputs changed', inputsChanged)
```
pmuens: Looks like we can remove this.

src/utils/components/executeComponent.js
```diff
@@ -58,4 +87,98 @@ const executeComponent = async (
...
+
+  /* Has state, run update flow if inputsChanged */
+  if (inputsChanged) {
+    context.log(`${context.type}: ○ Updating`)
+    /* Get critical values from inputTypes */
+    const criticalValues = Object.keys(inputTypes).filter((inputKey) => {
+      const inputData = inputTypes[inputKey]
+      return inputData.critical
+    })
+    // console.log('criticalValues', criticalValues)
```
pmuens: Looks like we can remove this.

src/utils/components/executeComponent.js
```diff
@@ -58,4 +87,98 @@ const executeComponent = async (
...
+
+    /* Then check if a critical value has changed */
+    const criticalValueChanged = criticalValues.some((r) => { // eslint-disable-line
+      return componentData.keys.includes(r)
+    })
+    // console.log('criticalValueChanged', criticalValueChanged)
```
pmuens: Looks like we can remove this.

src/utils/components/getComponentFunctions.js
```diff
@@ -1,9 +1,13 @@
 const path = require('path')
+const fileExistsSync = require('../fs/fileExistsSync')
 
 const getComponentFunctions = (componentRoot) => {
   let fns = {}
   try {
-    fns = require(componentRoot) // eslint-disable-line global-require, import/no-dynamic-require
+    const indexPath = path.resolve(componentRoot, 'index.js')
+    if (fileExistsSync(indexPath)) {
```
pmuens: This can be accomplished with an async call which is much more efficient (we should avoid sync calls since they've significantly slowed down Serverless Framework v1).

Something like `if (!await fileExists(indexPath)) {}` should do the trick :+1:

registry/github-webhook/index.js
brianneisler: +1 to this. The upper case confuses me cause I always think it's a function that's meant to be instantiated with `new`.

src/utils/components/compareInputsToState.js
brianneisler: +1

src/utils/components/getComponentFunctions.js
brianneisler: +1

registry/github-webhook/index.js
DavidWells: delete is reserved keyword. The api is `Create`, `Update`, `Delete` semantics

src/utils/components/getComponentFunctions.js
DavidWells: I tried await but got unhandled promise warnings

```
(node:79970) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: Cannot find module '/Users/davidwells/Dropbox/Projects/serverless/scratch/components-eslam/examples/github-webhook-example'
(node:79970) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

For whatever reason sync fs lookup was the only way I could get this working correctly

registry/github-webhook/github.js
DavidWells: I would have but that does not work with multiline comments

DavidWells:
image

```bash
components # will show all commands help
components -h # will show all commands help
components --help # will show all commands help

components webhookFunction # will show all commands for component 'webhookFunction'
components webhookFunction -h # will show all commands for component 'webhookFunction'
components webhookFunction --help # will show all commands for component 'webhookFunction'
```

rupakg:
The ASCII art is too over powering. Could be a little subtle. The `help` functionality looks great.

bin/components
```diff
@@ -1,12 +1,13 @@
 #!/usr/bin/env node
-const minimist = require('minimist')
-
+const parseArgs = require('minimist')
+const { dissoc } = require('ramda')
```
pmuens: This needs to be updated to

```js
const dissoc = require('ramda').dissoc
```

since the `bin` file is not transpiled by Babel.

registry/aws-lambda/commands/invoke.js
```diff
@@ -0,0 +1,36 @@
+const AWS = require('aws-sdk')
+
+const LAMBDA_CONFIG = process.env.AWS_REGION
+  ? { apiVersion: '2015-03-31' }
+  : { apiVersion: '2015-03-31', region: 'us-east-1' }
+
+function invoke(inputs, state, options) {
```
pmuens: This could be an `async function`

registry/github-webhook/index.js
```diff
@@ -1,211 +1,112 @@
...
+      }
 
-  const params = {
-    name: 'web',
-    owner: gh.owner,
-    repo: gh.name,
-    config: {
-      url: payloadUrl,
-      content_type: 'json'
+      return getWebhook({
```
pmuens: Can we rewrite this with `async` / `await` here to make it more readable?

registry/github-webhook/index.js
```diff
@@ -1,211 +1,112 @@
...
 }
 
 module.exports = {
-  deploy,
-  remove
+  commands,
+  Create,
+  Update,
+  Delete,
+  myCustomFunctionToDoStuff: (inputs, state, options) => {
```
pmuens: Do we need to keep this here? Shouldn't we move this to the docs or somewhere else?

registry/github-webhook/serverless.yml
```diff
@@ -7,11 +7,19 @@ license: Apache-2.0
 author: "Serverless, Inc. <hello@serverless.com> (https://serverless.com)"
 repository: "github:serverless/components"
 
+commands:
+  yolo:
```
pmuens: This should be updated or moved to the docs (I'd vote to move this to the docs).

src/run.js
```diff
@@ -36,25 +39,105 @@ const run = async (command, options) => {
...
-    await executeGraph(graph, components, stateFile, archive, command, options, false)
-    // run the "info" command on every component after a successful deployment
-    if (command === 'deploy') {
-      // NOTE: need to re-build the graph here since we're mutating it in "executeGraph"
-      // TODO: we should refactor this code later on
-      // eslint-disable-next-line no-shadow
-      const graph = await buildGraph(componentsToUse, orphanedComponents, 'info')
-      await executeGraph(graph, components, stateFile, archive, 'info', options, false)
+      // Do the CLI command from individual component
+      await doCommand(command, options, availableCommands, {
```
pmuens: Maybe `runCommand` rather than `doCommand`?

src/run.js
```diff
@@ -83,4 +166,223 @@ const run = async (command, options) => {
...
 
+async function doCommand(cmd, options, availableCommands, data) {
+  const parts = cmd.split(' ')
+
+  const componentKey = parts[0]
+  const commandData = availableCommands[componentKey]
+  const currentCommand = parts[1]
+
+  if (!commandData) {
+    console.log(`No ${componentKey} component found`) // eslint-disable-line
```
pmuens: Eslam introduced a `log` method which can be used here (and everywhere else).

src/run.test.js
```diff
@@ -34,37 +34,12 @@ beforeEach(() => {
 })
 
 describe('#run()', () => {
-  it('should run the command on the graph', async () => {
```
pmuens: A similar test should be kept.

src/run.test.js
```diff
@@ -34,37 +34,12 @@ beforeEach(() => {
...
-    expect(utils.writeStateFile).toHaveBeenCalled()
-    expect(utils.errorReporter).toHaveBeenCalled()
-  })
-
   it('should report any errors to Sentry while still writing the state to disk', async () => {
-    utils.executeGraph.mockImplementation(() => Promise.reject(new Error('something went wrong')))
-
-    await expect(run('some-command', {})).rejects.toThrow('something went wrong')
+    utils.executeGraph.mockImplementation(() =>
+      // eslint-disable-line
```
pmuens: Won't this break the return statement of the `mockImplementation`?

src/run.test.js
```diff
@@ -34,37 +34,12 @@ beforeEach(() => {
...
+      Promise.reject(new Error('something went wrong')))
 
+    await expect(run('deploy', {}))
```
pmuens: Initially we've expected it to throw an error. Does the current implementation suppress the error?

src/utils/components/executeComponent.js
```diff
@@ -58,4 +86,114 @@ const executeComponent = async (
   return component
 }
 
+async function defaultRemove(inputs, context, component) {
+  const fns = component.fns
+  // const inputTypes = component.inputTypes
```
pmuens: This should be removed.

src/utils/components/executeComponent.test.js
```diff
@@ -105,45 +105,6 @@ describe('#executeComponent()', () => {
     expect(res.outputs).toEqual({ result: 'deployed' })
   })
 
-  it('should skip the command execution if the command is not available', async () => {
```
pmuens: A test like that should be kept.

src/utils/components/getComponentFunctions.js
```diff
@@ -1,9 +1,17 @@
+const fs = require('fs')
 const path = require('path')
 
 const getComponentFunctions = (componentRoot) => {
   let fns = {}
   try {
-    fns = require(componentRoot) // eslint-disable-line global-require, import/no-dynamic-require
+    if (componentRoot && typeof componentRoot === 'string') {
+      const indexPath = path.resolve(componentRoot, 'index.js')
+      if (fileExistsSync(indexPath)) {
```
pmuens: Sync calls shouldn't be used here.

src/utils/components/getComponentFunctions.js
```diff
@@ -15,8 +23,19 @@ const getComponentFunctions = (componentRoot) => {
...
       }
     }
-  } // eslint-disable-line no-empty
+    // Do not swallow component code errors
+    throw error
+  }
   return fns
 }
 
+function fileExistsSync(filePath) {
```
pmuens: There's an `async` version in the `@serverless/utils` package which can be used instead.

src/utils/components/getComponentFunctions.test.js
```diff
@@ -1,14 +0,0 @@
-const { join } = require('path')
```
pmuens: Tests for that file should be kept.

tests/integration/simple.test.js
```diff
@@ -156,66 +156,6 @@ describe('Integration Test - Simple', () => {
       expect(myFunction.instanceId).not.toBeFalsy()
     })
 
-    it('should invoke the "function" component with CLI options', async () => {
```
pmuens: A test similar to this should be kept.

src/run.js
DavidWells: context doesn't exist here. Where is the log method?

pmuens: https://github.com/serverless/components/blob/master/src/utils/log.js

eahefnawy:
@ganeshvlrk yep that's the plan. Not sure about closing the PR though. I see no immediate need to do so.

@brianneisler what you think?

ganeshvlrk:
@eahefnawy I'd like for us to aim to not have PRs open for a long time without a clear action plan. Since we have decided to put this on the back burner, lets close out the PR. We can raise a new one when we get back to this. Can you please do the same tomorrow.

eahefnawy:
@ganeshvlrk sounds good 👍

PR #173, opened 2018-04-27
Download Component from URL

eahefnawy:
create & deploy the following component:

```yml
type: my-project
version: 0.0.1

components:
  downloadedComponent:
    # Component package I created for testing. Includes IAM and lambda inside
    type: https://github.com/eahefnawy/component-url-test/archive/master.zip
    inputs:
      test: abc
```
- You should see a message stating the component is downloading, then it will run its nested components (IAM and Lambda).

- Try deploying again, the component will be used from the cache and won't be downloaded again.

### DIFF
--- a/src/utils/components/downloadComponent.js
+++ b/src/utils/components/downloadComponent.js
@@ -7,10 +7,6 @@ const log = require('../log')
 module.exports = async (url) => {
   const downloadedComponentRootPath = await getComponentRootPathFromUrl(url)
   await fse.ensureDirAsync(downloadedComponentRootPath)
-  const componentName = url.substr(url.lastIndexOf('/') + 1).slice(0, -4)
-  log(`Downloading component: ${componentName}`)
-  return download(
-    url, downloadedComponentRootPath,
-    { extract: true, strip: 1 }
-  )
+  log(`Downloading component from => ${url}`)
+  return download(url, downloadedComponentRootPath, { extract: true, strip: 1 })
 }
--- a/src/utils/components/downloadComponent.js
+++ b/src/utils/components/downloadComponent.js
@@ -1,7 +1,8 @@
 const download = require('download')
+const BbPromise = require('bluebird')
+const fse = BbPromise.promisifyAll(require('fs-extra'))
 const getComponentRootPathFromUrl = require('./getComponentRootPathFromUrl')
 const log = require('../log')
-const { fse } = require('../fs')
 
 module.exports = async (url) => {
   const downloadedComponentRootPath = await getComponentRootPathFromUrl(url)
--- a/src/utils/components/downloadComponent.js
+++ b/src/utils/components/downloadComponent.js
@@ -6,7 +6,6 @@ const { fse } = require('../fs')
 module.exports = async (url) => {
   const downloadedComponentRootPath = await getComponentRootPathFromUrl(url)
   await fse.ensureDirAsync(downloadedComponentRootPath)
-  const componentName = url.substr(url.lastIndexOf('/') + 1).slice(0, -4)
   log(`Downloading component => ${url}`)
   return download(
     url, downloadedComponentRootPath,
--- a/src/utils/components/downloadComponent.js
+++ b/src/utils/components/downloadComponent.js
@@ -7,7 +7,8 @@ const log = require('../log')
 module.exports = async (url) => {
   const downloadedComponentRootPath = await getComponentRootPathFromUrl(url)
   await fse.ensureDirAsync(downloadedComponentRootPath)
-  log(`Downloading component => ${url}`)
+  const componentName = url.substr(url.lastIndexOf('/') + 1).slice(0, -4)
+  log(`Downloading component: ${componentName}`)
   return download(
     url, downloadedComponentRootPath,
     { extract: true, strip: 1 }
--- a/src/utils/components/downloadComponent.js
+++ b/src/utils/components/downloadComponent.js
@@ -10,6 +10,6 @@ module.exports = async (url) => {
   log(`Downloading component => ${url}`)
   return download(
     url, downloadedComponentRootPath,
-    { extract: true, filename: componentName, strip: 1 }
+    { extract: true, strip: 1 }
   )
 }
--- a/src/utils/components/downloadComponent.js
+++ b/src/utils/components/downloadComponent.js
@@ -1,0 +1,15 @@
+const download = require('download')
+const getComponentRootPathFromUrl = require('./getComponentRootPathFromUrl')
+const log = require('../log')
+const { fse } = require('../fs')
+
+module.exports = async (url) => {
+  const downloadedComponentRootPath = await getComponentRootPathFromUrl(url)
+  await fse.ensureDirAsync(downloadedComponentRootPath)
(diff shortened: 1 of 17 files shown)


src/utils/components/getComponentFromUrl.test.js
```diff
@@ -0,0 +1,34 @@
...
+const { fse } = require('../fs')
+
+jest.mock('./downloadComponent')
+
+afterAll(() => {
+  jest.restoreAllMocks()
+})
+
+afterEach(() => {
+  downloadComponent.mockClear()
```
pmuens: Same could be done here.

src/utils/components/getComponentRootPathFromUrl.js
```diff
@@ -0,0 +1,9 @@
+const path = require('path')
+const getComponentsCachePath = require('./getComponentsCachePath')
+
+module.exports = async (url) => {
+  const componentsCachePath = await getComponentsCachePath()
+  const componentName = url.substr(url.lastIndexOf('/') + 1).slice(0, -4)
```
pmuens: This fixed slicing will fail if we support other formats such as `.tar.gz` or `.7zip`.

src/utils/components/getComponentsCachePath.test.js
```diff
@@ -0,0 +1,27 @@
...
+    ensureDirAsync: jest.fn()
+  }
+}))
+
+afterAll(() => {
+  jest.restoreAllMocks()
+})
+
+afterEach(() => {
+  fs.fse.ensureDirAsync.mockClear()
```
pmuens: Same here

src/utils/components/getComponentRootPathFromUrl.js
eahefnawy: yep you're right. Problem is components name could include dots. So how could we insure we parse a url like this to get the component name?

```
example.com/some.file.name.tar.gz
```

both the name and extension could have any number of dots

Furthermore, the compression package we use only support .tar and .zip anyway, and the story states that we need to support .tar and zip. So I think we can keep this as it is for now

eahefnawy:
Thanks for the review @pmuens ... I pushed some updates. Please take another look.

re: race condition: components are downloaded in parallel as they're unrelated. This shouldn't be an issue.

re: packaging: I tried it and the only error I got was about adding the version. I added that the packaging worked as expected.

src/utils/components/getComponentRootPathFromUrl.js
pmuens: Ok, that makes sense. Let's keep it as is.

eahefnawy:
oooh nono "Downloaded Component" message is coming from the deploy method of my test repo:

https://github.com/eahefnawy/component-url-test/blob/master/index.js#L3

So it doesn't really say "the component was downloaded". Just a test message for me to know that the deploy hook ran. My bad, should've made it more clear 😅

src/utils/components/getComponentRootPathFromUrl.js
manafount: I hate that I'm about to suggest RegEx for readability, but something like 

```
const regex = /([^/]+)((\.zip)|(\.tar)|(\.7zip)|(\.tar\.gz))$/ 
const componentName = url.match(regex)[1]
const compressionFormat = url.match(regex)[2] // if necessary for some other purpose...
```

might make it more obvious to contributors that only `.zip`, `.tar`, etc formats are supported, while letting you easily add more supported file formats later if you end up switching compression packages.

brianneisler: That said, for general download by url I don't know that this value is unique enough for a cache identifier. This could easily result in collisions in a situation like this...
```yml
type: my-project
version: 0.0.1

components:
  component1:
    type: https://my-website.com/component1/package.zip
  component2:
    type: https://my-website.com/component2/package.zip  # collides with component1
```
Perhaps we should hash the entire url and use that as the cache identifier? This then gives us a mechanism for cache busting as well by adding a query parameter to the url.

src/utils/components/downloadComponent.js
```diff
@@ -0,0 +1,15 @@
...
+const BbPromise = require('bluebird')
+const fse = BbPromise.promisifyAll(require('fs-extra'))
+const getComponentRootPathFromUrl = require('./getComponentRootPathFromUrl')
+const log = require('../log')
+
+module.exports = async (url) => {
+  const downloadedComponentRootPath = await getComponentRootPathFromUrl(url)
+  await fse.ensureDirAsync(downloadedComponentRootPath)
+  log(`Downloading component => ${url}`)
+  return download(
```
brianneisler: Is there any downside or adverse affects to having all component execute from a global directory like this?

eahefnawy: which directory are you talking about? the cache? This is where components are downloaded

src/utils/components/getComponentRootPath.js
```diff
@@ -1,10 +1,24 @@
...
+const semVerDefinition =
+  '(?:0|[1-9]\\d*)\\.(?:0|[1-9]\\d*)\\.(?:0|[1-9]\\d*)(?:-[\\da-z-]+(?:\\.[\\da-z-]+)*)?(?:\\+[\\da-z-]+(?:\\.[\\da-z-]+)*)?'
+const registryRegex = new RegExp(`^${typeRegexDefinition}@${semVerDefinition}$`)
+
+async function getComponentRootPath(type = null) {
   if (isNil(type)) {
     return process.cwd()
+  } else if (urlRegex({ exact: true }).test(type)) {
+    return getComponentFromUrl(type)
+  } else if (type.match(registryRegex)) {
```
pmuens: Why don't we use this case as the `else` case (if nothin else matches)? We could replace the `return` at the very bottom (`return path.resolve(type)`) with the logic here 🤔

This way we could get rid of the regexes above and simplify the code.

src/utils/components/getComponentsBucketRoot.js
```diff
@@ -0,0 +1,2 @@
+module.exports = () =>
+  'https://s3.amazonaws.com/serverless-components-registry/'
```
pmuens: Just wondering if we should use the `s3` URL direcly here or something like `registry.serverless.com` which is more generic and easier to modify / manage by us 🤔.

Maybe KISSing it is the best for this initial iteration...

src/utils/components/getComponentRootPath.js
eahefnawy: we want to keep both because:

- `aws-lambda@0.0.2` would download from the bucket
- `aws-lambda` would use the local registry

Eventually when we move the registry out of the core we could do what you're suggesting, but for now, we're just keeping that functionality.

eahefnawy: RE regex: The regex above is taken from the SDK repo. Imo it's more robust as it also validates for a valid semver, not simply "anything with `@`"

RE registry: updating the all registry components to download other sub components will require pumping their version and uploading them again to S3. and the only way to do that now is by pushing a tag, which should happen in a different PR in the next release. So we need to keep the reference to the core local registry for now otherwise things will break during this transition period.

PR #184, opened 2018-05-02
Upload Registry Components to S3

eahefnawy:
TODO

- [x] set api keys in travis


scripts/prepublish.js
```diff
@@ -1,21 +1,100 @@
...
+  sentryDSN: SENTRY_DSN,
   environment: 'production',
-  segmentWriteKey: process.env.SEGMENT_WRITE_KEY
-};
-(() => writeFile(trackingConfigFilePath, trackingConfig))()
+  segmentWriteKey: SEGMENT_WRITE_KEY
+}
+
+const uploadComponent = async (componentRoot) => {
+  const packageTempDirPath = path.join(os.tmpdir(), crypto.randomBytes(3).toString('hex'))
```
pmuens: There's also a [`getTmpDir`](https://github.com/serverless/utils/blob/master/src/fs/getTmpDir.js) function in utils you could use. This function already namespaces the tmp directories. Furthermore you could get rid of some `require` calls at the top.

scripts/prepublish.js
```diff
@@ -1,21 +1,100 @@
...
+  await fse.ensureDirAsync(packageTempDirPath)
+  const options = {
+    format: FORMAT,
+    path: packageTempDirPath,
+    componentRoot
+  }
+  const packagePath = await packageComponent(options)
+  const componentPackage = await fse.readFileAsync(packagePath)
```
pmuens: Not sure how large the file is, but using a `readStream` would greatly reduce memory consumption since you're not reading the whole file and keep it in memory. This is especially interesting when this code runs in a CI system where server resources are usually pretty sparse.

Changing that should be pretty straightforward. We did something similar in the Serverless Framework repo with this PR --> https://github.com/serverless/serverless/pull/3145

scripts/prepublish.js
```diff
@@ -1,21 +1,100 @@
...
+    const slsYml = await readFile(slsYmlFilePath)
+    const componentFileName = `${slsYml.type}@${slsYml.version}.${FORMAT}`
+    if (!s3Components.includes(componentFileName)
+      && !excludedComponents.includes(path.basename(componentRoot))) {
+      accum.push(componentRoot)
+    }
+    return accum
+  }, Promise.resolve([]))
+
+  return BbPromise.all(componentsToUpload.map(uploadComponent))
```
pmuens: You could use `BbPromise.map` which would make the `BbPromise.all` obsolete.

scripts/prepublish.js
```diff
@@ -1,21 +1,100 @@
...
+
+const getUploadedComponents = async () => (await s3.listObjectsV2({ Bucket: COMPONENTS_BUCKET })
+  .promise()).Contents.map((obj) => path.basename(obj.Key))
+
+const uploadComponents = async () => {
+  const componentsRoots = await getRegistryComponentsRoots()
+  const s3Components = await getUploadedComponents()
+
+  const componentsToUpload = await componentsRoots.reduce(async (accum, componentRoot) => {
+    accum = await Promise.resolve(accum)
```
pmuens: We shouldn't mix native and Bluebird promises here since those might introduce hard to debug race-conditions.

src/utils/components/packageComponent.js
```diff
@@ -8,7 +8,8 @@ const pack = require('../pack')
 
 module.exports = async (options) => {
   const format = options.format || 'zip'
-  const slsYmlFilePath = path.join(process.cwd(), 'serverless.yml')
+  const componentRoot = options.componentRoot || process.cwd()
```
pmuens: Can you add a test for the `cwd` case?

src/utils/getRegistryComponentsRoots.js
```diff
@@ -0,0 +1,10 @@
+const path = require('path')
```
pmuens: Can you add a test for this functionality?

scripts/prepublish.js
eahefnawy: ❤️

scripts/prepublish.js
eahefnawy: ❤️ ❤️ ❤️ ❤️

scripts/prepublish.js
eahefnawy: ❤️

scripts/prepublish.js
eahefnawy: 👍

src/utils/components/packageComponent.js
eahefnawy: we already have one for cwd. I'm guessing you mean componentRoot. I just pushed that 🙌

src/utils/getRegistryComponentsRoots.js
eahefnawy: 👍

src/utils/components/packageComponent.js
pmuens: Thanks! 👍

scripts/prepublish.js
```diff
@@ -1,21 +1,96 @@
...
-if (!process.env.SENTRY_DSN) {
-  throw new Error('SENTRY_DSN env var not set')
-}
+const SENTRY_DSN = process.env.SENTRY_DSN
+const SEGMENT_WRITE_KEY = process.env.SEGMENT_WRITE_KEY
+
+const COMPONENTS_BUCKET = process.env.COMPONENTS_BUCKET
+const COMPONENTS_BUCKET_REGION = process.env.COMPONENTS_BUCKET_REGION
+const COMPONENTS_BUCKET_API_KEY = process.env.COMPONENTS_BUCKET_API_KEY
+const COMPONENTS_BUCKET_API_SECRET = process.env.COMPONENTS_BUCKET_API_SECRET
```
pmuens: This could be further simplified with destructuring:

```js
const {
  COMPONENTS_BUCKET,
  COMPONENTS_BUCKET_REGION,
  COMPONENTS_BUCKET_API_KEY,
  COMPONENTS_BUCKET_API_SECRET
  // ...
} = process.env
```

scripts/prepublish.js
```diff
@@ -1,21 +1,96 @@
...
+    Body: fs.createReadStream(packagePath),
+    Bucket: COMPONENTS_BUCKET,
+    Key: path.basename(packagePath)
+  }
+
+  return s3.putObject(params).promise()
+}
+
+const getUploadedComponents = async () => (await s3.listObjectsV2({ Bucket: COMPONENTS_BUCKET })
+  .promise()).Contents.map((obj) => path.basename(obj.Key))
```
pmuens: Will this map work with an empty bucket? You could use Ramdas `map` here which won't throw if the to-be-mapped-array is empty.

src/utils/getRegistryComponentsRoots.js
```diff
@@ -0,0 +1,10 @@
...
+const BbPromise = require('bluebird')
+const fse = BbPromise.promisifyAll(require('fs-extra'))
+const getRegistryRoot = require('./getRegistryRoot')
+
+const getRegistryComponentsRoots = async () => (await fse.readdirAsync(getRegistryRoot()))
+  .filter(async (f) => (await fse.statAsync(path.join(getRegistryRoot(), f))).isDirectory())
```
pmuens: You could use https://github.com/serverless/utils/blob/master/src/fs/dirExists.js here.

scripts/prepublish.js
```diff
@@ -1,21 +1,96 @@
...
+
+const uploadComponents = async () => {
+  const componentsRoots = await getRegistryComponentsRoots()
+  const s3Components = await getUploadedComponents()
+
+  const componentsToUpload = await componentsRoots.reduce(async (accum, componentRoot) => {
```
pmuens: http://bluebirdjs.com/docs/api/promise.reduce.html could be used here as well.

scripts/prepublish.js
eahefnawy: yep. Tested it with an empty bucket

PR #192, opened 2018-05-09
Implement output type validation

pmuens:
## What has been implemented?

Implements the `outputTypes` YAML property which can be used to define and check component outputs.

Closes SC-94

## Steps to verify

Modify the `tests-integration-function-mock` to require some ouput types which are not present and run the `simple` integartion test or deploy the `tests/integration/simple` service.

You might also want to `deploy` and `remove` the `retail-app` to see that everything still works fine.

## Todos:

* [x] Write tests
* [x] Run Prettier


src/utils/components/validateTypes.js
```diff
@@ -0,0 +1,31 @@
...
+    if (!validation.valid) {
+      const component = componentId.split(':').pop()
+      const componentName = chalk.white(`"${component}"`)
+      const header = chalk.redBright.bold(`\nType error(s) in component ${componentName}:\n`)
+      errorMessages = append(header, errorMessages)
+      forEach((error) => {
+        const value = chalk.cyanBright(`"${error.value}"`)
+        const key = `${chalk.yellowBright(error.key)}`
+        const suppliedType = typeof error.value
+        const msg = `  - ${key} has \`${suppliedType}\` value of ${value} but expected the following: ${chalk.yellowBright(error.rule)} ${chalk.yellowBright(error.attr)}.\n`
```
eahefnawy: The message does not distinguish between inputTypes and outputTypes. What happens if an inputType and an outputType are identical? (name as both input and output)

scripts/doc-gen.js
```diff
@@ -10,22 +10,19 @@ const config = {
   transforms: {
```
eahefnawy: sweet addition 👍

src/utils/components/validateTypes.js
pmuens: Yes, good catch! I thought about exacrtly the same thing. The reason I implemented it that way was to make it re-usable w/o providing any special parameter to just trigger the "inputs" / "outputs" string concatenation.

But you're right. It can be confusing since `inputTypes` and `outputTypes` can be the same (and they are in some `serverless.yml` files).

scripts/doc-gen.js
pmuens: Thanks!

src/utils/components/validateTypes.js
brianneisler: @pmuens one way to solve this would be to have this function accept another function that’s responsible for generating the error messages to the user. You can then create a validateOutputTypes method that passes in an appropriate message generator and another validateInputTypes method that does the same.

pmuens: Thanks @brianneisler! 👍

pmuens: @eahefnawy @brianneisler just fixed this by introducing an optional `opts` object which includes a `prefix` property.

src/utils/components/getComponent.test.js
```diff
@@ -2,10 +2,18 @@ const path = require('path')
 const getComponent = require('./getComponent')
 
 describe('#getComponent()', () => {
+  const registryPath = path.join(__dirname, '..', '..', '..', 'registry')
```
eahefnawy: better use the `getRegistryRoot()` util

pmuens: Good idea 👍 Thanks!

pmuens:
@eahefnawy thanks for your review! 👍

I just rebased and updated the PR according to your comments.

scripts/doc-gen.js
brianneisler: Love it!

rupakg: this is sweet!

pmuens:
Thanks for the review @brianneisler !
You're right. We should document that (For some reason I removed the checkbox to add documentation from the PR description). I just pushed a commit which adds the documentation.

README.md
```diff
@@ -326,6 +319,23 @@ components:
...
+
+Output types are the description of the outputs your component returns. You supply those `outputTypes` in the component's `serverless.yml` file:
+
+```yaml
+type: my-component
+
+outputTypes:
+  firstName:
+    type: string
+    required: true # firstName will always be returned
```
brianneisler: @pmuens a required flag on outputs types doesn't really make sense to me. Whether or not a property is capable of being null DOES make sense, but that would be declared in the type declaration. I know we're using the same typing system under the hood for both inputs and outputs, so i don't think there's a code change required but it would be best not to mention this property in the docs for outputTypes.

pmuens: Yes, `required` for output types also felt odd to me (and is slightly mind-bending when thinking about it).

However I think there's a need to say "this output will be returned 100% and is not optional".

PR #195, opened 2018-05-10
Use Serverless Components as a library

pmuens:
## What has been implemented?

Implements the functionality to pull in Serverless Components as a library and use it programatically.

Closes #SC-265

### Pointing to a project path

Create a `workflow-with-path.js` file a new directory called `library-usage` in the `examples` directory:

```js
// examples/library-usage/workflow-with-path.js

const { join } = require('path')
const { log } = require('../../src/utils')
const { pkg, deploy, remove } = require('../../src')

const projectPath = join('..', '..', 'tests', 'integration', 'library-usage')

async function run() {
  log('Packaging service...')
  await pkg({ projectPath, path: servicePath })
  log('Deploying service...')
  await deploy({ projectPath })
  log('Re-deploying service...')
  await deploy({ projectPath })
  log('Removing service...')
  await remove({ projectPath })
}

run()
```

Run `node workflow-with-path.js`.

### Providing a `serverless.yml` file object

Create a `workflow-with-serverless-file-object.js` file a new directory called `library-usage` in the `examples` directory:

```js
// examples/library-usage/workflow-with-serverless-file-object.js

const { log } = require('../../src/utils')
const { pkg, deploy, remove } = require('../../src')

const serverlessFileObject = {
  type: 'my-app',
  version: '0.0.1',
  components: {
    myRole: {
      type: 'tests-integration-iam-mock',
      inputs: {
        name: 'my-role-name',
        service: 'my.function.service'
      }
    }
  }
}

async function run() {
  // NOTE: right now the cwd() is packaged. Not sure how this behavior is desired / if we
  // should support packaging when using a serverlessFileObject
  // log('Packaging service...')
  // await pkg({ path: process.cwd(), serverlessFileObject })
  log('Deploying service...')
  await deploy({ serverlessFileObject })
  log('Re-deploying service...')
  await deploy({ serverlessFileObject })
  log('Removing service...')
  await remove({ serverlessFileObject })
}

run()

```

Run `node workflow-with-serverless-file-object.js`.

## Todos:

* [x] Fix issue when re-deploying with projectPath
* [x] Write tests
* [x] Run Prettier
* [x] Add documentation

### DIFF
--- a/bin/components
+++ b/bin/components
@@ -8,10 +8,6 @@ const command = process.argv[2]
 const options = minimist(process.argv.slice(2))
 delete options._
 
-if (!options.projectPath) {
-  options.projectPath = process.cwd()
-}
-
 run(command, options).catch((error) => {
   console.error(error)
   process.exit(1)
--- a/bin/components
+++ b/bin/components
@@ -7,7 +7,12 @@ const { run } = require('../src/index')
 const command = process.argv[2]
 const options = minimist(process.argv.slice(2))
 delete options._
-run(process.cwd(), command, options).catch((error) => {
+
+if (!options.projectPath) {
+  options.projectPath = process.cwd()
+}
+
+run(command, options).catch((error) => {
   console.error(error)
   process.exit(1)
 })
--- a/bin/components
+++ b/bin/components
@@ -7,7 +7,7 @@ const { run } = require('../src/index')
 const command = process.argv[2]
 const options = minimist(process.argv.slice(2))
 delete options._
-run(command, options).catch((error) => {
+run(process.cwd(), command, options).catch((error) => {
   console.error(error)
   process.exit(1)
 })
--- a/README.md
+++ b/README.md
@@ -141,6 +141,10 @@ Also please do join the _Components_ channel on our public [Serverless-Contrib S
     * [deploy](#deploy)
     * [info](#info)
     * [remove](#remove)
+  * [Programmatic usage](#programmatic-usage)
+    * [deploy](#deploy)
+    * [package](#package)
+    * [remove](#remove)
   * [Component Docs](#component-docs)
     * [aws-apigateway](./registry/aws-apigateway)
     * [aws-cloudfront](./registry/aws-cloudfront)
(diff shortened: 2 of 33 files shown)


eahefnawy:
Also, I'm curious to know why you're referring to the cwd as a service rather than a component? 😅 ... cause I've been battling with the difference when I think about v1/v2 integration.

eahefnawy:
Oh, also, I think it's very important to return the values of all provisioned component outputs to be consumed programmatically outside of serverless.yml. The usecase for this is again v1 integration, where we will use those values to populate the rest of v1 service (ie. if we reference a component output from within the resources section)

Maybe also returning the state? Don't have a usecase for that though, but thought it might be useful

bin/components
```diff
@@ -6,7 +6,7 @@ const { run } = require('../src/index')
 const command = process.argv[2]
 const options = minimist(process.argv.slice(2))
 delete options._
-run(command, options).catch((error) => {
+run(process.cwd(), command, options).catch((error) => {
```
brianneisler: @pmuens i think the cwd could be passed in as an option and if not provided, then assume it is the cwd. That seems like the simplest approach

pmuens: Thanks for pointing me in that direction @brianneisler 👍 - Yes, that sounds like an easier, more reasonable change. I'll update the code accordingly.

bin/components
```diff
@@ -6,6 +6,11 @@ const { run } = require('../src/index')
 const command = process.argv[2]
 const options = minimist(process.argv.slice(2))
 delete options._
+
+if (!options.projectPath) {
```
brianneisler: @pmuens it might make more sense to put this defaulting into the run method. It's a bit nicer when all options are actually optional when using the module programmatically

pmuens: Yes, that totally makes sense :+1:

pmuens:
Than kyou very much for your review @brianneisler ! 👍

I just updated the PR to address the usages of `process.cwd()` and added the `projectPath` to the context so that components can use it as well.

I removed the functionality to default to `process.cwd()` in `getComponentRootPath` since AFAIK it's not used anymore.

pmuens:
Hmm 🤔

There seems to be an issue with re-deploying when providing a `projectPath`. My first guess is that it's related to the state file retrieval. I'll look into that...

pmuens:
@brianneisler @eahefnawy I just updated this PR and replaced the option parameters to add support the `serverlessFileObject` option.

Tests are almost passing. Only the packaging seems to be broken right now. I'll look into it and fix it.

Aside: I updated the PR description which now includes an example on how to use the new `serverlessFileObject` option.

pmuens:
@eahefnawy @brianneisler I just rebased and updated the PR so that a `projectPath` and the `serverlessFileObject` options are now supported.

All tests have been fixed and the steps to manually test this PR were added to the PR description.

pmuens:
I just updated it according to your comments.

> Could we also add docs for passing the serverless config object?
Nice! Good catch!

> Also did you confirm that this object can contain serverless variables and they will be resolved?
This implementation doesn't touch too many parts of the core. The only thing which was changed is the way things are loaded right in the beginning of the `run` method even before executing the graph. Therefore the main parts such as Variable resolution should still work. I haven't tested it though.

> Other than that, we are now only exposing deploy, remove and package, so all other utils are not exposed. Correct?

Yes. In our case it's `deploy`, `pkg` and `remove`. `package` is a reserved JavaScript keyword.

pmuens:
> @pmuens oh, so we are not exposing a package function?!

Our `package` function is called `pkg` since `package` is a reserved keyword in JavaScript. IMHO I don't like it and think it's "ugly". Do you have any idea how we could name it instead? Or should we keep `pkg`?

eahefnawy:
@pmuens ugh! naming things! `pkg` is fine 👍

pmuens:
> @pmuens ugh! naming things! pkg is fine 👍

Awesomesauce! Thanks @eahefnawy 👍 Merging...

PR #199, opened 2018-05-12
Add AWS Lambda runtime

laardee:
## What has been implemented?

This PR adds runtime option to AWS Lambda function component.

## Steps to verify

* Deploy Lambda function with different runtimes, e.g. nodejs6.10 and nodejs8.10

## Todos:

* [x] Write tests
* [x] Run Prettier


registry/aws-lambda/serverless.yml
```diff
@@ -26,6 +26,14 @@ inputTypes:
     displayName: Function timeout
     description: The function execution time at which Lambda should terminate the function
     example: 3
+  runtime:
+    type: string
+    required: true
```
rupakg: Add a ‘default’ attribute with a value of nodejs8.10

laardee: Done ✅ In which stage the default value is applied if it is not defined to the component?

laardee:
There are now tests for the create (https://github.com/serverless/components/pull/199/files#diff-9008fd066156821dd60d09863ef6f926R94) and the update (https://github.com/serverless/components/pull/199/files#diff-9008fd066156821dd60d09863ef6f926R174). Both of them have `nodejs6.10` runtime in the inputs and the test checks that `createLambda` and `updateLambda` calls the mock with correct runtime and other parameters.

PR #213, opened 2018-05-17
Extend linting rules

pmuens:
## What has been implemented?

In https://github.com/serverless/components/pull/205 we've replaced the "airbnb" liniting rules with "prettier" linting rules to tightly integrate both, ESLint as well as Prettier.

However some really helpful rules from the airbnb plugin are missing.

I added those via this PR. Might add more via other PRs if something else is missing.

## Todos:

* [x] Run Prettier

### DIFF
--- a/registry/s3-website-config/index.js
+++ b/registry/s3-website-config/index.js
@@ -8,7 +8,7 @@ const setBucketForWebsiteConfig = async ({
   rootBucketName,
   indexDocument,
   errorDocument,
-  redirectBucketName,
+  redirectBucketName, // eslint-disable-line no-unused-vars
   redirectToHostName // eslint-disable-line no-unused-vars
 }) => {
   const params = {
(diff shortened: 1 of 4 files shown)


registry/s3-website-config/index.js
```diff
@@ -27,8 +27,8 @@ const setBucketForWebsiteConfig = async ({
 }
 
 const setBucketForRedirection = async ({
-  rootBucketName,
-  indexDocument,
+  rootBucketName, // eslint-disable-line no-unused-vars
+  indexDocument, // eslint-disable-line no-unused-vars
```
eahefnawy: why are those variables unused? Why do we have this ESlint exception?

pmuens:
Thanks for the review @brianneisler 👍

I think this is GTM. We could address the component-related linting issues in another PR (if necessary).

PR #214, opened 2018-05-17
Add .env support for yml file

### LINKED ISSUE #200: Feature Request: Auto-load .env file
Could we auto-load `.env` files in the root of the parent component?  This would allow users to specify environment variables in their `serverless.yml` without having to save them to their bash shell. 
 Otherwise, the user has to create custom deployment logic.

ebisbe:
Load environment variables with dotenv

Closes #200

* [x] Write tests
* [x] Run Prettier


package.json
```diff
@@ -34,6 +34,8 @@
     "bluebird": "^3.5.1",
     "chalk": "^2.3.2",
     "concordance": "^3.0.0",
+    "crypto-extra": "^0.4.0",
```
brianneisler: Looks like the crypto-extra dependency may have regressed. I don't think we need this dep anymore.

tests/integration/env/serverless.yml
```diff
@@ -0,0 +1,8 @@
+type: envLoad
```
brianneisler: we've been trying to be consistent about naming conventions with component types. Could we rename this to `env-load` also, could we change the directory name to match the name of the component?

tests/integration/envLoad.test.js
```diff
@@ -0,0 +1,40 @@
...
+const BbPromise = require('bluebird')
+
+const fsp = BbPromise.promisifyAll(fse)
+const cpp = BbPromise.promisifyAll(cp)
+
+async function removeStateFiles(stateFiles) {
+  return BbPromise.map(stateFiles, (stateFile) => fsp.removeAsync(stateFile))
+}
+
+describe('Integration Test - Simple', () => {
```
brianneisler: Looks like copy pasta. Can we update the name to match "Integration Test - Env Load". Also, can we update the name of this file to be consistent with the other tests `env-load.test.js`. Thanks!

brianneisler:
Looks great! Thanks @ebisbe!

ac360:
@ebisbe This is a great improvement.  Thank you for your contribution  👍

pmuens:
Great stuff! Thanks for implementing @ebisbe 👍 💯

PR #223, opened 2018-05-25
Added setOutputs function to context

eahefnawy:
## What has been implemented?
Added setOutput method to the context instead of returning outputs directly.

Closes SC-115

## Steps to verify

Just run the basic example and to make sure it's still working.

## Todos:

* [x] Write tests
* [x] Run Prettier
* [x] Update all component

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,7 +134,7 @@ Also please do join the _Components_ channel on our public [Serverless-Contrib S
 * [Creating Components](#creating-components)
   * [Basic setup](#basic-setup)
   * [`serverless.yml`](#serverless.yml)
-  * [`index.js`](#index.js)
+  * [`index.js`](#indexjs)
   * [Testing](#testing)
 * [Docs](#docs)
   * [CLI Usage](#cli-usage)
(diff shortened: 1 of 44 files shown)


eahefnawy:
@ganeshvlrk @brianneisler had some final additions to add. Not sure if he got to it yet.

ganeshvlrk:
@brianneisler Seems like there is some additional testing which need to do here. When you have a chance, can you please outline what else is pending with this PR?

laardee:
@eahefnawy Following syntax doesn't seem to work with Node 4 and Jest version it is using when it runs the tests; Jest v22 cannot be used with Node 4, and it falls back to v21 while installing the dependencies. 
[src/utils/components/executeComponent.test.js line 168](https://github.com/serverless/components/blob/e3fb2b440700f774454578410fae301bf423bb70/src/utils/components/executeComponent.test.js#L168)
```js
await expect(
  executeComponent(componentId, components, stateFile, archive, command, options)
).rejects.toThrow('Type error(s)')
```

Is Node 4 support required, as it's LTS status is already "end-of-life"? If so, something like this could be used instead.
https://github.com/serverless/components/compare/set-outputs...laardee:set-outputs
```js
let res
try {
  res = await executeComponent(componentId, components, stateFile, archive, command, options)
} catch (error) {
  expect(error.message).toContain('Type error(s)')
}
expect(res).toBeUndefined()
```

brianneisler:
Given the type system is likely to replace outputs I am closing this PR in favor of the work we are doing there.

PR #229, opened 2018-06-26
Add Babel watch script to the registry

laardee:
## What has been implemented?

This PR adds Babel watch script to the registry. It should ease the component development when running test deployments with development version.

## Steps to verify

* In registry directory, run `npm run watch` to watch all components
* `npm run watch -- aws-s3-bucket eventgateway` to only watch specified components

## Todos:

* [ ] Write tests
* [x] Run Prettier

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,7 +27,7 @@
     "@octokit/rest": {
       "version": "14.0.9",
       "resolved": "https://registry.npmjs.org/@octokit/rest/-/rest-14.0.9.tgz",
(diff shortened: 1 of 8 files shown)


laardee:
@pmuens, something like this https://github.com/serverless/components/pull/229/commits/abf7ba316df2c4f067416bbf1a2a4493e709b9c0? Now the registry watch is using the watch.js script, but do you happen to know if it is possible to config babel so that it has multiple sources with multiple destinations? Like each component has src and dist directories. I tried few different syntaxes but without luck... `babel src **/src --out-dir **/dist ....` was promising, but it the output went so same directory 😞

Issue #241, opened 2018-08-12
Component Diff'ing


ac360:
Component authors need a way to understand what `inputs` have changed by users, so they can write better deployment logic.

Creating this thread to start that conversation.

ac360:
This idea was proposed by a few stakeholders:

```
A. The component author is expected to export two functions when he wants to handle the diffing by himself (that’s what we currently have in place):

- `deploy` - this function *creates* and *updates* the required infra based on the state. The component author writes the diffing logic himself
- `remove` - this function removes the entire infra, based on the state

B. When the component author wants to let the framework handle input diffing, he is expected to export three alternative functions:

- `create` - this function creates the required infra from scratch
- `update` - this function updates the required infra *without teardown*
- `delete` - this function deletes the entire infra

Then in his `serverless.yml`, the component author specifies which inputs trigger a `teardown` when changed, and which just runs an update. Something like this:

```inputTypes:
  name:
    teardown: true # changing the lambda name for example requires deleting + creating from scratch
  memory:
    teardown: false # changing the memory should just update```

The framework then checks what inputs have changed by comparing to the saved inputs from the previous state, and can determine the right action based on this `teardown` property (`update` or `delete + create`). Although this is now harder to do since the introduction of `saveState()`, because we can’t guarantee that the previous inputs have been saved to state. So the component author either needs to save the inputs to state himself, or the framework could automatically do that.

That’s basically it at a high level! I noticed that most providers/infra can follow this pattern. With this implementation, component authors don’t need to have all those state specific if statements in the deploy function in most cases. When this is not flexible enough for them, they can revert to the `deploy + remove` pair instead.
```

ac360:
Interesting idea.  Here is my feedback:

As a component author, my immediate reaction is maintaining these additional methods feel like more work.  Perhaps, it may not be a lot of extra work since the logic written can be used for all methods, but it kinda feels that way and that's enough to be a component author disincentive.  `deploy` and `remove` are simple.

As a component author, I'm not sure the CUD option is possible in a handful of my use-cases.  I am building abstract components over infra components (e.g. the Function or Subscription component) and the Framework cannot make CUD decisions because it's dependent on how the infra works which the user has configured in the abstract component.  It's possible many component implementations cannot use this method.  We need to learn more.

Because of increased complexity and the fact that it can't be used for some components, I believe we should wait and learn more before implementing the above suggestion.

The community will teach us about components.  Keeping it general and simple for now is the best way for users to tell us what they want.

[...]

PR #242, opened 2018-08-13
Diff'ing POC

ac360:
## What has been implemented?

Created a POC of diff'ing functionality to solve #241.  This POC adds a `context.changes` property which contains an object showing changed inputs.  It works for deeply nested inputs and even supports array modifications.  This attempts to model an API which component authors can use to write more sophisticated deployment logic and depend upon for the long-term.

## Steps to verify

* Deploy a component
* Update its inputs and look at `context.changes`

Sample:

```js
[
     { change: 'update', 
       path: [ 'timeout' ], 
       previous: 12, 
       current: 10 },
     { change: 'create',
       path: [ 'foo' ],
       previous: null,
       current: 'bar' }, 
] 
```

## Todos:

* [x] Write tests
* [x] Run Prettier

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,7 @@ dist
 .eslintcache
 yarn.lock
 default.etcd
+workspace
 
 .serverlessrc
 .serverless
--- a/examples/basic/serverless.yml
+++ b/examples/basic/serverless.yml
@@ -3,7 +3,7 @@ version: 0.0.1
 
 components:
   myFunction:
-    type: aws-lambda@0.2.0
+    type: aws-lambda
     inputs:
       memory: 512
       timeout: 10
@@ -11,6 +11,6 @@ components:
       role:
         arn: ${myRole.arn}
   myRole:
-    type: aws-iam-role@0.2.0
+    type: aws-iam-role
     inputs:
       service: lambda.amazonaws.com
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,9 +19,9 @@
       "integrity": "sha1-lr3GtD4TSCASumaRsQGEktOWIsw=",
       "dev": true,
       "requires": {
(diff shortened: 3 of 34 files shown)


eahefnawy:
Thanks for this @ac360 🙌 ... You've put a lot of work into this!

Two questions I have:

1. I see you've added `getPreviousInputs` function, how do you ensure that the previous inputs have been saved to state?
2. Is there an example on how this new `updates` function could be used by component authors to simplify writing components? At first glance it seems that component authors still need to check this new method for what has changed, and react accordingly, rather than having the framework call the appropriate function on its own. I could still imagine some if statements that could be omitted.

ac360:
Hi @eahefnawy, thanks for the questions:

> I see you've added getPreviousInputs function, how do you ensure that the previous inputs have been saved to state?

They already are.  Every Component in `state.json` has an `inputs` object that shows the previous inputs.

> Is there an example on how this new updates function could be used by component authors to simplify writing components? At first glance it seems that component authors still need to check this new method for what has changed, and react accordingly, rather than having the framework call the appropriate function on its own. I could still imagine some if statements that could be omitted.

In this scenario, no extra Component methods are needed.  Instead, we just keep `deploy` and `remove`.  Component CUD methods currently don't seem logical to me (as described in #241).  The big difference is that authors would use diffs to assist their `deploy` logic, as opposed to having to remove/recreate everything or write your own diff logic, which is how components are being designed currently.

I am personally dying for this feature.  I'm confident I can reduce code and write components that deploy faster and more safely with this.  If we could merge this and just call it experimental, we can test it out right away.

eahefnawy:
ah yikes! Test coverage drop is preventing the merge.

pmuens:
Closing this PR because of an upcoming repository update...

PR #253, opened 2018-09-04
Google Cloud Function fixes / enhancements

pmuens:
This PRs brings the Google Cloud Function component to a stable state. Furthermore tests and docs are added.

Here's an example project which can be used to test it:

```yml
# serverless.yml

type: gcf-test

components:
  googleCloudFunction:
    type: google-cloud-function
    inputs:
      name: gcf-component-test
      projectId: my-project-id-1234
      keyFilename: ~/.gcloud/my-key.json
      entryPoint: http
      sourceCodePath: ${self.path}
      # NOTE: only one event type is supported per function
      # httpsTrigger:
      #   url: my-url
      eventTrigger:
        eventType: providers/cloud.pubsub/eventTypes/topic.publish
        resource: projects/my-project-id-1234/topics/my-topic
      environmentVariables:
        foo: bar
```

```javascript
// index.js

'use strict'

exports.http = (request, response) => {
  response.status(200).send('Hello World!')
}

exports.event = (event, callback) => {
  callback()
}
```


pmuens:
@eahefnawy the `keyFilename` points to your credentials file.

ac360:
I strongly recommend not using `keyfile` and custom credential methods for components.  This makes components more complicated to get started with.  This precedent will be followed by others, just like how the AWS Lambda component did this and now the GCF component is doing this as well.  This will continue to cause a bigger mess.

pmuens:
I agree, that `keyfile` looks like an antipattern. I've added it to the list of improvements we should tackle next up (full writeup can be found [in this issue](https://github.com/serverless/components/issues/258)).

I'd suggest that we merge this PR ASAP and work on the problems described in #258 (which includes removing the `keyfile` as a source for credentials). This PR serves the purpose of finalizing, testing and refactoring the component. Include more fixes would bload it which makes the review process harder.

registry/google-cloud-function/src/index.test.js
```diff
@@ -0,0 +1,396 @@
...
+afterAll(() => {
+  jest.restoreAllMocks()
+})
+
+afterEach(() => {
+  jest.clearAllMocks()
+  googleapis.mocks.mockCreate.mockClear()
+  googleapis.mocks.mockGet.mockClear()
+  googleapis.mocks.mockDelete.mockClear()
+  googleapis.mocks.mockPatch.mockClear()
```
eahefnawy: shouldn't clearAllMocks be enough?!

pmuens: No, AFAIK that won't work since those mocks are nested and live in `googleapis` mock. I already tried that with the `aws-lambda` component where we do the same and if we switch to `clearAllMocks` it won't clear all the mocks.

pmuens:
Thanks @eahefnawy! 👍

PR #255, opened 2018-09-06
Docker image component

pmuens:
Implements a Docker image component which makes it possible to build and publish Docker images.

The following config can be used to test the `build` / `deploy` and `remove` commands:

```yml
# serverless.yml

type: docker-image-test

components:
  docker:
    type: docker-image
    inputs:
      dockerfilePath: ${self.path}/Dockerfile
      contextPath: ${self.path}
      tag: jdoe/my-image:latest
      registryUrl: https://my-registry.com # leave this empty if the image should be pushed to the public Docker registry
      username: jdoe
      password: s0m3p455w0rd
```

```
# Dockerfile

FROM hello-world
```

You can run `components build` / `components deploy` to build / push the Docker image. Once built and pushed you can remove the Docker image via `components remove`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,3 +10,7 @@ RUN apt-get update -y \
   && apt-get upgrade -y \
   && apt-get install -y nodejs build-essential \
   && npm install npm@latest -g
+
+# install Docker
+RUN curl -fsSL get.docker.com -o get-docker.sh
+RUN sh get-docker.sh
--- a/README.md
+++ b/README.md
@@ -158,6 +158,7 @@ Also please do join the _Components_ channel on our public [Serverless-Contrib S
     * [aws-sns-platform-endpoint](./registry/aws-sns-platform-endpoint)
     * [aws-sns-subscription](./registry/aws-sns-subscription)
     * [aws-sns-topic](./registry/aws-sns-topic)
+    * [docker-image](./registry/docker-image)
     * [eventgateway](./registry/eventgateway)
     * [github-webhook](./registry/github-webhook)
     * [github-webhook-aws](./registry/github-webhook-aws)
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,3 +12,5 @@ services:
       - .:/app
       - ~/.aws/:/root/.aws
       - ~/.gitconfig:/root/.gitconfig
+      # run Docker in Docker...
+      - /var/run/docker.sock:/var/run/docker.sock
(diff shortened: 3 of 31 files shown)


registry/docker-image-assembly/serverless.yml
```diff
@@ -0,0 +1,48 @@
...
+  tag:
+    type: string
+    required: true
+    displayName: Tag
+    description: The tag which should be used to tag the Docker image
+    example: jdoe/my-image:latest
+  registryUrl:
+    type: string
+    displayName: Registry URL
+    description: The Docker registry URL which should be used to push the image (leave empty if you want to push the image to the public Docker registry)
```
eahefnawy: shouldn't there be a default here?

registry/docker-image-assembly/serverless.yml
```diff
@@ -0,0 +1,48 @@
...
+    example: ./Dockerfile
+  contextPath:
+    type: string
+    required: true
+    displayName: Build context
+    description: The build context path which is used when building the Docker image
+    example: .
+  tag:
+    type: string
+    required: true
```
eahefnawy: oh is a tag required?!

registry/docker-image-assembly/src/utils/checkDockerSetup.test.js
```diff
@@ -0,0 +1,46 @@
...
+
+describe('#checkDockerSetup()', () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+  })
+
+  it('should throw if Docker is not installed', async () => {
+    isDockerInstalled.mockImplementation(() => false)
+    // NOTE: docker can't be running if it's not installed
+    // isDockerRunning.mockImplementation(() => false)
```
eahefnawy: should this comment be removed?

registry/docker-image-assembly/serverless.yml
pmuens: That's a good point 👍 I have added the default in the code, but putting it here makes a lot more sense. I'll update it...

registry/docker-image-assembly/serverless.yml
pmuens: Yes, the tag must be set. Otherwise pushing it to the registry is not possible...

registry/docker-image-assembly/src/utils/checkDockerSetup.test.js
pmuens: Yes, we can remove it. I kept it as a companion code snippet for the `NOTE:` comment above.

pmuens:
@eahefnawy @raeesbhatti thanks for the review and the comments!

I've rebased the PR (since I merged another PR which introduced linting enhancements) and addressed all the review comments. Just tested this PR again and it works fine with the provided config in the PR description :+1:

PR #277, opened 2018-09-12
AWS ECS Service support

raeesbhatti:
## What has been implemented?

Closes #XXXXX

## Steps to verify

Deploy an app with following serverless.yml:
```
type: my-application
version: 0.0.2

components:
  myAwsEcsFargateService:
    type: aws-fargate
    inputs:
      cpu: '256'
      memory: '512'
      serviceName: myAwsEcsFargateService
      containerDefinitions:
      - name: hello
        essential: true
        image: nginx:1.7.9
        portMappings:
        - containerPort: 80
          hostPort: 80
          protocol: tcp
      desiredCount: 1
      exposePublically: true
```

## Todos:

* [x] Write tests
* [x] Run Prettier


registry/aws-ecs-service/src/index.js
```diff
@@ -0,0 +1,83 @@
...
+
+const deploy = async (inputs, context) => {
+  const { state } = context
+
+  if (state.serviceName && inputs.serviceName !== state.serviceName) {
+    context.log('Change to ECS service name requires replacement. Making one now...')
+    await remove({}, context)
+  }
+
+  const { services } = await ecs.describeServices({ services: [inputs.serviceName] }).promise()
```
pmuens: How does this behave when we have to paginate due to lots of services? Will this ever be a problem because we're filtering based on `serviceName` here?

src/utils/logging/log.js
```diff
@@ -1,7 +1,7 @@
-module.exports = (message) => {
+module.exports = (...params) => {
   if (!process.env.CI) {
     try {
-      process.stdout.write(`${message}\n`)
```
pmuens: AFAIK there was a reason that this is `process.stdout.write`. Not sure if @brianneisler or @eahefnawy committed that back then... 🤔

registry/aws-ecs-taskdefinition/serverless.yml
```diff
@@ -0,0 +1,82 @@
+type: aws-ecs-taskdefinition
```
pmuens: Personal preference, but should this be called `taskdefinition` or `task-definition`?

registry/aws-ecs-taskdefinition/serverless.yml
```diff
@@ -0,0 +1,82 @@
...
+repository: "github:serverless/components"
+
+inputTypes:
+  family:
+    type: string
+    required: true
+    example: "sleep360"
+  cpu:
+    type: integer
+    required: false
```
pmuens: You can remove `required: false` because it defaults to `false`.

registry/aws-ecs-service/src/index.js
raeesbhatti: We're only requesting one service (`services: [inputs.serviceName]`), so, we wont face that scenario.

brianneisler: Is there ever a case where there can be more than one service with the same name?

raeesbhatti: On the ECS, no. Service name must be unique, it doesn't allow you to create multiple services with same name. So, I don't see this request failing because of naming conflicts.

src/utils/logging/log.js
raeesbhatti: If we look at the `console` implementation in NodeJS, it also uses `process.stdout` under the hood. So, using `console.log` writes effectively in the same way. The reason I prefer `console.log` is that I will print out the contents of JSON objects for example, where as `process.stdout.write` will simply print `[object Object]`.

raeesbhatti: Ref: https://nodejs.org/api/console.html#console_new_console_stdout_stderr_ignoreerrors

registry/aws-ecs-service/src/index.js
brianneisler: ok, what i would recommend then is to allow this value to be set but allow it to be optional. If the value is not supplied in an input, we should default it to a string value built using `context.instanceId`

brianneisler:
@raeesbhatti a note regarding the drop in codecov. It's likely this is happening due to this branch being behind master. If you merge master into this branch it might bring the code coverage back up to where it was before.

pmuens:
Closing this PR because of an upcoming repository update...

PR #281, opened 2018-09-12
Adds Amazon VPC Component

### LINKED ISSUE #266: AWS VPC component
# Description

Users may need to create a VPC in AWS, whether for their Lambda functions or for Fargate containers.

[AWS SDK for JS documentation on CreateVPC call](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/EC2.html#createVpc-property)

## inputTypes

`cidrBlock` - `string`
`amazonProvidedIpv6CidrBlock` - `boolean`
`instanceTenancy` - `string`. Choices: `default`, `dedicated` or `host`.

## outputTypes

`vpcId` - `string`

laardee:
## What has been implemented?

Closes #266 

## Steps to verify

Setup without any inputs should create a VPC with CIDR '10.0.0.0/16', amazonProvidedIpv6CidrBlock 'false' and instanceTenancy default.

```yaml
type: vpc-service

components:
  myAwsVpc:
    type: aws-vpc
```

Following setup should create a VPC with CIDR '10.0.0.0/20', amazonProvidedIpv6CidrBlock 'true' and instanceTenancy "dedicated".

```yaml
type: vpc-service

components:
  myAwsVpc:
    type: aws-vpc
    inputs:
      cidrBlock: 10.0.0.0/20
      amazonProvidedIpv6CidrBlock: true
      instanceTenancy: dedicated
```

## Todos:

* [x] Write tests
* [x] Run Prettier


registry/aws-vpc/src/index.js
```diff
@@ -0,0 +1,66 @@
...
+    return { vpcId: state.vpcId }
+  }
+  // any changes to vpc requires replacement
+  if (!isEmpty(state) && !compareStateAndInputs(state, inputs)) {
+    context.log(`Changes to existing VPC requires replacement`)
+    await remove(state, context)
+  }
+  context.log(`Creating a VPC`)
+  const { Vpc } = await ec2
+    .createVpc({
```
brianneisler: @laardee, in the event that the state is lost but the VPC has already been created, will that basically result in a second VPC in the same CidrBlock?

laardee: @brianneisler yes, and the new VPC details are then written to the state. Same way than with the other components that don't have any name or other unique parameters which prevent duplicates. 

In general, we could tag the AWS components with the `serviceId` and `instanceId` component and use that for the cases where we want to check wether something exists or not. We could great a resource group based on the `serviceId` tag also. Then if the state file gets lost, you then have some visibility what have been created and can delete those by hand. Or we could provide method for that.

brianneisler: @laardee cool, that makes sense and I like the idea of using tags to help solve this problem in the future. Will add an issue to track that idea.

brianneisler: https://github.com/serverless/components/issues/282

Issue #275, opened 2018-09-11
AWS EC2 Security Group Ingress component


alexdebrie:
Relates to #274 

# Description

Users may need to create a security group such as for their Lambda functions or Fargate containers. This component allows them to add an ingress rule to their security group.

[AWS SDK for JS documentation on AuthorizeSecurityGroupIngress call](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/EC2.html#authorizeSecurityGroupIngress-property)

## inputTypes

`cidrIp` - `string`
`fromPort` - `integer`
`groupId` - `string`
`groupName` - `string`
`ipPermissions` - `Array<map>`
*  `fromPort` - `integer`
*  `ipProtocol` - `string`
*  `ipRanges` - `Array<map>`
   * `cidrIp` - `string`
   * `description` - `string`
* `ipv6Ranges` - `Array<map>`
   * `cidrIpv6` - `string`
   * `description` - `string`
* `prefixListIds` - `Array<map>`
   * `prefixListId` - `string`
   * `description` - `string`
* `toPort` - `integer`
* `userIdGroupPairs` - `Array<map>`
   * `description` - `string`
   * `groupId` - `string`
   * `groupName` - `string`
   * `peeringStatus` - `string`
   * `userId` - `string`
   * `vpcId` - `string`
   * `vpcPeeringConnectionId` - `string`
`ipProtocol` - `string`
`sourceSecurityGroupName` - `string`
`sourceSecurityGroupOwnerId` - `string`
`toPort` - `integer`

## outputTypes

_None_

Issue #274, opened 2018-09-11
AWS EC2 Security Group component


alexdebrie:
# Description

Users may need to create a security group such as for their Lambda functions or Fargate containers.

[AWS SDK for JS documentation on CreateSecurityGroup call](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/EC2.html#createSecurityGroup-property)

## inputTypes

`description` - `string`
`groupName` - `string`
`vpcId` - `string`

## outputTypes

`groupId` - `string`

NwaforAugustine321:
@alexdebrie I want know if this issue is still open for me to work on it. Thanks and look forward to hearing from you.

alexdebrie:
@NwaforAugustine321 I'm not sure! I'm not involved with this project anymore.

Issue #272, opened 2018-09-11
AWS Route Table Association component


alexdebrie:
Relates to #266, #267 and #270 

# Description

Users may need to create a VPC in AWS, whether for their Lambda functions or for Fargate containers. This component lets a user associate a route table with a subnet.

[AWS SDK for JS documentation on AssociateRouteTable call](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/EC2.html#associateRouteTable-property)

## inputTypes

`routeTableId` - `string`
`subnetId` - `string`

## outputTypes

`associationId` - `string`

laardee:
https://github.com/serverless/components/pull/285 propably should be merged before starting this.

KrishnaSolo:
Hi is this still open? I would love to help out if so!

Issue #271, opened 2018-09-11
AWS VPC Route component


alexdebrie:
Related to #266 and #270 

# Description

Users may need to create a VPC in AWS, whether for their Lambda functions or for Fargate containers. This component allows them to create a route in their VPC's route table.

[AWS SDK for JS documentation on CreateRoute call](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/EC2.html#createRoute-property)

## inputTypes

`destinationCidrBlock` - `string`
`destinationIpv6CidrBlock` - `string`
`egressOnlyInternetGatewayId` - `string`
`gatewayId` - `string`
`instanceId` - `string`
`natGatewayId` - `string`
`networkInterfaceId` - `string`
`routeTableId` - `string`
`vpcPeeringConnectionId` - `string`

## outputTypes

None

laardee:
https://github.com/serverless/components/pull/285 propably should be merged before starting this.

PR #285, opened 2018-09-16
Adds AWS subnet, AWS Internet Gateway, and AWS VPC Gateway Attachment components

### LINKED ISSUE #267: AWS VPC Subnet component
Related to #266 

# Description

Users may need to create a VPC in AWS, whether for their Lambda functions or for Fargate containers. This component allows them to specify a subnet within the VPC.

[AWS SDK for JS documentation on CreateSubnet call](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/EC2.html#createSubnet-property)

## inputTypes

`availabilityZone` - `string`
`cidrBlock` - `string`
`ipv6CidrBlock` - `string`
`vpcId` - `string`

## outputTypes

`subnetId` - `string`

### LINKED ISSUE #268: AWS Internet Gateway component
Related to #266

# Description

Users may need to create a VPC in AWS, whether for their Lambda functions or for Fargate containers. This component allows them to create an internet gateway for the VPC.

[AWS SDK for JS documentation on CreateInternetGateway call](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/EC2.html#createInternetGateway-property)

## inputTypes

_None_

## outputTypes

`internetGatewayId` - `string`

### LINKED ISSUE #269: AWS VPC Gateway Attachment component
Related to #266 and #268

# Description

Users may need to create a VPC in AWS, whether for their Lambda functions or for Fargate containers. This component allows them to attach a created internet gateway to a VPC.

[AWS SDK for JS documentation on AttachInternetGateway call](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/EC2.html#attachInternetGateway-property)

## inputTypes

`internetGatewayId` - `string`
`vpcId` - `string`

## outputTypes

`internetGatewayId` - `string`

laardee:
## What has been implemented?

Closes #267 
Closes #268 
Closes #269 

## Steps to verify

Subnet component
```yaml
components:
  myAwsVpc:
    type: aws-vpc
    inputs:
      cidrBlock: 10.0.0.0/16
  myAwsSubnet1:
    type: aws-subnet
    inputs:
      vpcId: ${myAwsVpc.vpcId}
      availabilityZone: us-east-1a
      cidrBlock: 10.0.0.0/24
  myAwsSubnet2:
    type: aws-subnet
    inputs:
      vpcId: ${myAwsVpc.vpcId}
      availabilityZone: us-east-1b
      cidrBlock: 10.0.1.0/24
```

Internet Gateway and VPC Gateway Attachment as separate components
```yaml
components:
  myAwsVpc:
    type: aws-vpc
    inputs:
      cidrBlock: 10.0.0.0/16
  myAwsIntergergateway:
    type: aws-internetgateway
  myAwsIntergergatewayattachment:
    type: aws-vpcgatewayattachment
    inputs:
      vpcId: ${myAwsVpc.vpcId}
      internetGatewayId: ${myAwsIntergergateway.internetGatewayId}
```

If `vpcId` is defined in Internet Gateway component, it will load the VPC Gateway Attachment component and use it to attach IGW to VPC.

```yaml
components:
  myAwsVpc:
    type: aws-vpc
    inputs:
      cidrBlock: 10.0.0.0/16
  myAwsIntergergateway:
    type: aws-internetgateway
    inputs:
      vpcId: ${myAwsVpc.vpcId}
```

## Gotchas

As I was expecting, I ran into a dependency issue with VPC. That is why I took more than one issue to this same PR. 

VPC has strong dependencies to resources that are connected to it. That means that it cannot be removed before dependencies are removed. Changing VPC parameters requires replacement, which leads to a situation where VPC with old parameters should be removed, but it still has e.g. subnet attached to it.

How this is now solved, is that I call the removal but I don't wait for it, so the removal goes to end of the execution queue in the thread. Now the new VPC is created before the old one is removed and creation continues to the subnet which removes the old resource because the `vpcId` is changed and creates a new one. It is kind of how the CloudFormation handles this also.

Then the dependencies. The Components removal order is somehow wrong what comes to the VPC. It tries to remove VPC before subnet or Internet Gateway, which causes VPC removal to fail. I added `waitForDependenciesToBeRemoved` function to poll the describe APIs to find out when it is ok to remove the VPC. Issues #270, #271, #272, #274, and #275 has similar dependency to VPC.

## Todos:

* [x] Write tests
* [x] Run Prettier


registry/aws-vpcgatewayattachment/src/index.js
```diff
@@ -0,0 +1,66 @@
...
+  return equals(statePick, inputsPick)
+}
+
+const deploy = async (inputs, context) => {
+  const { state } = context
+  if (compareStateAndInputs(state, inputs, ['internetGatewayId', 'vpcId'])) {
+    return { internetGatewayId: state.internetGatewayId }
+  }
+  context.log('Creating Internet Gateway Attachment')
+  await ec2
```
brianneisler: I think this works for the programmatic use case above where this component is directly used in the `aws-internetgateway` component. However, I think this case will fail if this component is declared in an actual components block when either the `internetGatewayId` or the `vpcId` input changes. This method should probably check if either the `internetGatewayId` and/or the `vpcId` has changed compared to state. If changed, call `remove` before proceeding with the call to `attachInternetGateway`.

laardee: Oh, I totally missed that one. Yes, it needs to be removed, otherwise, VPC cannot be removed either when it changes.

PR #297, opened 2018-10-01
Google cloud integration

pmuens:
Adds the types for the Google Cloud integration.

- [x] `GoogleCloudProvider`
- [x] `GoogleCloudFunctionCompute`
- [x] `GoogleCloudFunction`


src/types/GoogleCloudProvider/serverless.yml
```diff
@@ -0,0 +1,36 @@
...
+# locationId: ${inputs.locationId}
+
+inputTypes:
+  clientEmail:
+    type: String
+    required: true
+    displayName: Client E-Mail
+    description: The client E-Mail which can be found in the keyfile.json
+    example: acme@example-project.iam.gserviceaccount.com
+  privateKey:
```
grayside: Explicit configuration of the private key here seems a bit risky. The recommended approach is to set the environment variable GOOGLE_APPLICATION_CREDENTIALS to the local path of your keyfile.

At that point google.auth.getClient() shown in the [samples in the client README](https://github.com/googleapis/google-api-nodejs-client) will automatically determine the correct authentication mechanism based on the credential type and environment. (You can look deeper at those details in https://github.com/googleapis/google-auth-library-nodejs)

For this component, maybe there could simple be a parameter to the path of the keyfile which can be set as an env var in JS before the authentication process is triggered?

pmuens: Hey @grayside thanks for looking into this! :+1:

Yes, that's a good a valid point. We've struggled a lot with credentials in the past and wanted to come up with a uniform solution all providers should adhere to so we've decided to pass in the credentials via input. This way everything is explicit and consistent.

AFAIK @brianneisler is planning to add another piece in here which will grab the `privateKey` from the environment variables and pass them into `inputs`. This way we get the best of both worlds.

src/types/GoogleCloudFunction/serverless.yml
```diff
@@ -6,4 +6,119 @@ implements:
...
+entryPoint: ${inputs.entryPoint}
+environmentVariables: ${inputs.environmentVariables}
+labels: ${inputs.labels}
+maxInstances: ${inputs.maxInstances}
+network: ${inputs.network}
+sourceArchiveUrl: ${inputs.sourceArchiveUrl}
+sourceRepository: ${inputs.sourceRepository}
+sourceUploadUrl: ${inputs.sourceUploadUrl}
+eventTrigger: ${inputs.eventTrigger}
+httpsTrigger: ${inputs.httpsTrigger}
```
grayside: Making sure I understand, inputTypes are defined below, but in order to pass them to the component they need to be mapped to property names above?

src/types/GoogleCloudFunctionCompute/serverless.yml
```diff
@@ -0,0 +1,40 @@
...
+    type: Provider
+    default: ${context.provider}
+    displayName: Provider
+    description: The reference to the Cloud provider type
+    example: ${context.provider}
+  runtime:
+    type: String
+    default: nodejs
+    displayName: Runtime
+    description: The functions runtime
```
grayside: It might help if runtime is generally documented more like an enum of the available options. nodejs6, nodejs8, python37.

src/types/GoogleCloudFunction/utils/pack.js
```diff
@@ -0,0 +1,72 @@
...
+const { join } = require('path')
+const crypto = require('crypto')
+const os = require('os')
+const archiver = require('archiver')
+const { readFile, writeFile } = require('@serverless/utils')
+
+async function pack(code, tempPath) {
+  tempPath = tempPath || os.tmpdir()
+
+  const tmpStatgingDir = join(tempPath, 'staging')
```
ace-n: Nit: typo (`statging` --> `staging`)

src/types/GoogleCloudFunctionCompute/shims/shim.js
```diff
@@ -0,0 +1,62 @@
+exports.shim = (...args) => {
+  // load the "uniform function"
+  const fileName = process.env.SERVERLESS_HANDLER.split('.')[0]
```
ace-n: Teeny/FYI nit: put this in a variable instead of duplicating it.

eahefnawy:
Sorry @pmuens 😅 ... we now have core support for google creds though 👌

PR #301, opened 2018-10-05
Rest API changes

raeesbhatti:
## What has been implemented?

Closes #XXXXX

## Steps to verify

* Run X
* ...

## Todos:

* [ ] Write tests
* [ ] Run Prettier

### DIFF
--- a/src/types/AwsApigateway/index.js
+++ b/src/types/AwsApigateway/index.js
@@ -1,0 +1,122 @@
+const { equals } = require('ramda')
+const { pick } = require('@serverless/utils')
+const { getSwaggerDefinition, generateUrl, generateUrls } = require('./utils')
+
+const inputsProps = ['provider', 'name', 'roleArn', 'routes']
+
+const deleteApi = async (APIGateway, params) => {
+  const { id } = params
+
+  await APIGateway.deleteRestApi({
+    restApiId: id
+  }).promise()
+  const outputs = {
+    id: null,
+    url: null,
+    urls: null
+  }
+  return outputs
+}
+
+const createApi = async (APIGateway, params) => {
+  const { name, roleArn, routes } = params
+
+  const swagger = getSwaggerDefinition(name, roleArn, routes)
+  const json = JSON.stringify(swagger)
+
+  const res = await APIGateway.importRestApi({
+    body: Buffer.from(json, 'utf8')
+  }).promise()
+
+  await APIGateway.createDeployment({
+    restApiId: res.id,
+    stageName: 'dev'
+  }).promise()
+
+  const url = generateUrl(res.id)
+  const urls = generateUrls(routes, res.id)
+
+  const outputs = {
+    id: res.id,
+    url,
+    urls
+  }
+  return outputs
+}
+
+const updateApi = async (APIGateway, params) => {
+  const { name, roleArn, routes, id } = params
+
+  const swagger = getSwaggerDefinition(name, roleArn, routes)
+  const json = JSON.stringify(swagger)
+
+  await APIGateway.putRestApi({
+    restApiId: id,
+    body: Buffer.from(json, 'utf8')
+  }).promise()
+
+  await APIGateway.createDeployment({
+    restApiId: id,
+    stageName: 'dev'
+  }).promise()
+
+  const url = generateUrl(id)
+  const urls = generateUrls(routes, id)
+
+  const outputs = {
+    id,
+    url,
+    urls
+  }
+  return outputs
+}
+
+export default {
+  async deploy(prevInstance, context) {
+    const APIGateway = new this.provider.sdk.APIGateway({
+      region: process.env.AWS_DEFAULT_REGION || 'us-east-1'
+    })
+    const inputs = pick(inputsProps, this)
+    const noChanges =
+      inputs.name === context.state.name &&
+      inputs.roleArn === context.state.roleArn &&
+      equals(inputs.routes, context.state.routes)
+
+    let outputs
+    if (noChanges) {
+      outputs = context.state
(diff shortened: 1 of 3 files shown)


src/types/EventGatewayProvider/index.js
```diff
@@ -0,0 +1,17 @@
+const EventGateway = require('@serverless/event-gateway-sdk')
+
+const AwsProvider = {
+  construct({ apiUrl, configurationUrl, apiKey, space }) {
+    this.sdk = new EventGateway({
```
brianneisler: These values could be variables. You will want to wait to build the SDK until it is needed (in the `getSdk` method) and then ensure you resolve the values before doing so. You can resolve all shallow properties and values of an object easily using the `all` method from `@serverless/utils`

src/types/EventGatewayProvider/serverless.yml
```diff
@@ -0,0 +1,25 @@
+name: EventGatewayProvider
+extends: Component
```
brianneisler: This should extend `Provider` not `Component`

src/types/EventGatewayProvider/serverless.yml
```diff
@@ -0,0 +1,25 @@
...
+main: ./index.js
+version: 0.2.0
+
+description: "Event Gateway Provider Serverless Component."
+license: Apache-2.0
+author: "Serverless, Inc. <hello@serverless.com> (https://serverless.com)"
+repository: "github:serverless/components"
+
```
brianneisler: Be sure to set the credentials object here using your inputs.

registry/AwsApigateway/serverless.yml
```diff
@@ -0,0 +1,37 @@
+name: AwsApiGateway
+extends: Component
+main: ./dist/index.js
+version: 0.2.0
```
brianneisler: Version `0.3.0` please. This will be the next version of components.

registry/AwsApigateway/src/index.test.js
```diff
@@ -0,0 +1,109 @@
+const ComponentType = require('./index')
```
brianneisler: Please use `import` here

registry/AwsApigateway/src/index.test.js
```diff
@@ -0,0 +1,109 @@
...
+// todo mock timers
+jest.setTimeout(16000)
+
+class SuperClass {
+  constructor(inputs) {
+    this.inputs = inputs
+  }
+}
+
+const SuperContext = {
```
brianneisler: A better way of testing this is to construct a context using the `createContext` method and then load the type instead of trying to reconstruct it here. This seems error prone.

registry/RestApi/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "@serverless-components/restapi",
+  "version": "0.3.0",
+  "description": "",
+  "main": "index.js",
```
brianneisler: main is at `dist/index.js`

registry/AwsApiGateway/src/utils.js
```diff
@@ -0,0 +1,237 @@
...
+      const apiGatewayIntegration = getApiGatewayIntegration(roleArn, uri, isCorsEnabled)
+      const securityDefinition = getSecurityDefinition(
+        methodObject.authorizer,
+        name,
+        undefined,
+        normalizedPath,
+        normalizedMethod
+      )
+      const defaultResponses = getDefaultResponses(isCorsEnabled)
+      updatedMethods = set(lensPath([normalizedMethod]), apiGatewayIntegration, updatedMethods)
```
brianneisler: I don't think the lensPath is needed here. Can simply use the array
```
updatedMethods = set([normalizedMethod], apiGatewayIntegration, updatedMethods)
```

registry/AwsApiGateway/src/index.js
```diff
@@ -0,0 +1,160 @@
...
+        inputs.name === state.name &&
+        (inputs.role && state.role && inputs.role.arn === state.role.arn) &&
+        equals(inputs.routes, state.routes)
+
+      let outputs
+      if (noChanges) {
+        outputs = state
+      } else if (inputs.name && !state.name) {
+        context.log(`Creating API Gateway: "${inputs.name}"`)
+        outputs = await createApi(APIGateway, inputs, this.provider.inputs.region)
```
brianneisler: I don't think I would access another instances inputs like this. This seems like a fragile implementation. The contract between an instance and the outside world is what it makes available on its properties. 

This same value is already placed  onto the provider as a property for the AwsProvider. You should be able to access the region like this
```
this.provider.region
```

registry/AwsApiGateway/src/index.js
```diff
@@ -0,0 +1,160 @@
...
+      } else {
+        context.log(`Updating API Gateway: "${inputs.name}"`)
+        outputs = await updateApi(
+          APIGateway,
+          {
+            ...inputs,
+            id: state.id,
+            baseUrl: state.baseUrl
+          },
+          this.provider.inputs.region
```
brianneisler:
```
this.provider.region
```

registry/AwsApiGateway/src/index.js
```diff
@@ -0,0 +1,160 @@
...
+                }
+                return methodAcc
+              },
+              [],
+              pathObj
+            )
+          )
+        },
+        [],
+        this.inputs.routes || {}
```
brianneisler: You'll want to resolve the routes here on this check. Variable will still pass this check even if the variable points to something that is undefined.

```
resolve(this.inputs.routes) || {}
```

alternatively you could use the `or` method from utils
```
or(this.inputs.routes, {})
```

PR #315, opened 2018-10-23
Implementation of info command

raeesbhatti:
## What has been implemented?

Closes #XXXXX

## Steps to verify

* Run X
* ...

## Todos:

* [ ] Write tests
* [x] Run Prettier


src/plugins/Info/index.js
```diff
@@ -1,18 +1,69 @@
...
-    context.log('TODO: implement info command')
+const printArray = (arr, log, level = 1) => {
+  arr.forEach((item) => {
+    const { type, data } = item
+    let { title } = item
+    if (!title && data.title) {
+      title = data.title
+    }
+    log(`\n|- ${`  `.repeat(level)}${title} - ${type}`)
+    if (Array.isArray(data)) {
```
brianneisler: Use `isArray` from utils. I suggest using the utils methods for these kinds of things because it enables a layer of indirection. If we want to change how these methods work we can then do so in one place and it takes effect everywhere instead of having to go back and manually update all calls.

src/plugins/Info/index.js
```diff
@@ -1,18 +1,69 @@
...
+    }
+  })
+}
 
-    return context
+const printObj = (obj, log, level = 1) => {
+  if (!obj) {
+    return
+  }
+  if (Object.keys(obj).length === 3 && obj.title && obj.type && obj.data) {
```
brianneisler: `keys` method from utils

src/plugins/Info/index.js
```diff
@@ -1,18 +1,69 @@
...
+    if (Array.isArray(obj.data)) {
```
brianneisler: `isArray` method from utils

src/plugins/Info/index.js
```diff
@@ -1,18 +1,69 @@
...
+      printArray(obj.data, log, level + 1)
+    } else {
+      printObj(obj.data, log, level)
+    }
+  } else {
+    Object.entries(obj).forEach(([key, val]) => {
```
brianneisler: this can be simplified with the `forEach` method from utils
```
forEach((val, key) => ..., obj)
```

src/plugins/Info/index.js
```diff
@@ -1,18 +1,69 @@
...
+      const space = '  '
+      if (isObject(val)) {
```
brianneisler: Array is an object and returns true for `isObject`. Do you mean to check for that here?

PR #341, opened 2018-11-07
Overhaul AwsDynamoDb component

pmuens:
This PR adds a complete overhaul of the `AwsDynamoDb` component.

/cc @brianneisler @plfx @ganeshvlrk 

It removes all the unnecessary clutter and dependencies (which caused errors such as the infamous `DTrace` one) and adds a 100% tested codebase which complies with the new type system.

**Note:** The DynamoDB table creation and removal process is async on AWS end. However we immediately respond that the table was created / removed. I choose to implement it that way for the time-being since it's a little bit faster and the DynamoDB table is usually created in a more complex environment (and shouldn't take that long if you don't overprovision it). We might want to introduce a monitoring functionality soon to fix this issue. Code for smth. like that was alreay written and can be found in https://github.com/serverless/components/pull/297

Here's a `serverless.yml` you can use to test this PR:

```yml
name: ComponentsTest
extends: Component

credentials:
  type: EnvironmentBasedConfiguration
  inputs:
    variables:
      - AWS_ACCESS_KEY_ID
      - AWS_SECRET_ACCESS_KEY
      - AWS_SECURITY_TOKEN
      - AWS_SESSION_TOKEN

providers:
  aws:
    type: AwsProvider
    inputs:
      credentials:
        accessKeyId: ${this.credentials.values.AWS_ACCESS_KEY_ID}
        secretAccessKey: ${this.credentials.values.AWS_SECRET_ACCESS_KEY}
        securityToken: ${this.credentials.values.AWS_SECURITY_TOKEN}
        sessionToken: ${this.credentials.values.AWS_SESSION_TOKEN}
      region: us-east-1

components:
  testComponent:
    type: AwsDynamoDb
    inputs:
      provider: ${this.providers.aws}
      tableName: my-dynamodb-table
      attributeDefinitions:
        - AttributeName: id
          AttributeType: S
      keySchema:
        - AttributeName: id
          KeyType: HASH
      provisionedThroughput:
        ReadCapacityUnits: 1
        WriteCapacityUnits: 1
```


pmuens:
Sweet. Thanks for the review @dschep 👍

registry/AwsDynamoDB/src/index.test.js
```diff
@@ -0,0 +1,124 @@
...
+      provider: await context.construct(AwsProvider, {}),
+      tableName: 'updated-test-table'
+    })
+    nextAwsDynamoDb = await context.defineComponent(nextAwsDynamoDb, prevAwsDynamoDb)
+    nextAwsDynamoDb = resolveComponentEvaluables(nextAwsDynamoDb)
+
+    const result = nextAwsDynamoDb.shouldDeploy(prevAwsDynamoDb, context)
+
+    expect(result).toBe('replace')
+  })
```
brianneisler: @pmuens needs a test that ensures the properties instances are the same when no deploy is performed.

brianneisler: @pmuens also needs a test for idempotency on remove and idempotency on deploy

registry/AwsDynamoDB/src/utils/createTable.js
```diff
@@ -0,0 +1,31 @@
...
+    localSecondaryIndexes,
+    sseSpecification,
+    streamSpecification
+  } = properties
+
+  const SDK = provider.getSdk()
+  const ddb = new SDK.DynamoDB()
+
+  return ddb
+    .createTable({
```
brianneisler: Does this fail if a table with the same name already exists (idempotency)

registry/AwsDynamoDB/src/utils/deleteTable.js
```diff
@@ -0,0 +1,14 @@
+const deleteTable = async (properties) => {
+  const { provider, tableName } = properties
+
+  const SDK = provider.getSdk()
+  const ddb = new SDK.DynamoDB()
+
+  return ddb
+    .deleteTable({
```
brianneisler: does this fail if the table has already been deleted? (idempotency)

registry/AwsDynamoDB/src/utils/createTable.js
pmuens: This function alone will throw an error when using it in isolation.

However since this is used in combination with `shouldDeploy` in `index.js` it won't be called a second time if the table was already created in a previous deployment.

registry/AwsDynamoDB/src/index.test.js
pmuens: Looking into that...

registry/AwsDynamoDB/src/utils/deleteTable.js
pmuens: Same reasoning as above ☝️ (in isolation it fails, but the state management of core will ensure that this code isn't triggered twice).

pmuens:
@brianneisler just updated the code so that it can deal with an `ResourceNotFoundException` error on removals (--> idempotency).

I haven't added any specific handling for the case where the table already exists on removal because I think that this should still throw an error. Can update this easily if we should still throw an error, though.

pmuens:
Yes, good question @dschep 👍

All the SDK params are supported in this implementation so you should be able to configure it via `inputs` (if it's possible).

Looks like the `DeletionPolicy` configg is CloudFormation specific --> https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-deletionpolicy.html

registry/AwsDynamoDB/src/utils/createTable.js
brianneisler: @pmuens you can't guarantee that a table has not been created between the time that shouldDeploy runs and this method runs (simultaneous deployment running in parallel). Still have to wrap it in the correct error handling to cover these edge cases.

brianneisler:
@dschep good point, yes it would be good to add that functionality to this component. However, i think we might break that out into a separate feature that can be added to any component.

registry/AwsDynamoDB/src/index.test.js
pmuens: Just added some tests for that...

registry/AwsDynamoDB/src/utils/createTable.js
pmuens: That's a good point (love how the comment involves the nature of distributed systems) and I see that, but wouldn't it make more sense to fail rather than adapting to the fact that a table with the name was already created and moving on?

What if e.g. there's a table called `billings-prod` and I "accidentally" name my table `billings-prod`. IMHO deployment should absolutely fail here saying that such a table already exists rather than proceeding and letting the application write to `billings-prod` (and potentially causing issues with the state in the production data).

pmuens:
Hey @brianneisler @dschep  thanks for the reviews! :+1:

I just updated the PR to address all the issues.

@brianneisler not sure if I'm following https://github.com/serverless/components/pull/341#discussion_r231953652. I added a comment with my perspective on that below (maybe I misunderstand something here...).

pmuens:
@brianneisler I just rebased the PR so that it won't conflict with the base branch.

Except for `deploy` idempotency (see reasoning [here](https://github.com/serverless/components/pull/341#discussion_r232948105)) I've addressed all the changes you've requested. Since I'm the author I'm hesitant to push the button. Feel free to push the merge button if you think that this PR is good to go 👍

registry/AwsDynamoDB/src/utils/createTable.js
brianneisler: @pmuens the issue is you can't distinguish between that and a case where your deploy was interrupted (think power plug pulled) after the SDK call was made to create a table but before your app had a chance to save the state. 

On the next attempt to deploy, your app would now be hung because the table already exists but your state doesn't think it exists and you would continually get an error when trying to deploy. This is exactly why it's so important to treat both deploy and remove idempotently. The assumption should always be that the state declared in serverless.yml is the state we are trying to get to and if anything exists in the inf already we should be adjusting it to match what we want (regardless of what our state tells us). 

In your example, i think there is a different solution to offer protections of that sort. For instance, we could add tags to inf that connect it to a specific application. If a user is trying to deploy infrastructure and collides with another piece of inf that already exists, we can check to the tag to see if it's part of the same app, if not, we can consider it an error, if it is, then the deployment is free to modify it.

registry/AwsDynamoDB/src/index.js
```diff
@@ -1,345 +1,54 @@
...
-    )
-  },
-
-  async destroy(prevInstance, context) {
-    if (!prevInstance.tables || prevInstance.tables.length === 0) {
-      return {}
+      if (
+        prevInstance.tableName !== this.tableName ||
+        not(equals(prevInstance.attributeDefinitions, this.attributeDefinitions)) ||
+        not(equals(prevInstance.keySchema, this.keySchema)) ||
```
brianneisler: @pmuens i'm not sure all of these properties require replacement. We seem to be missing an update portion to this component. Mainly that seems pretty important given a replacement will result in wiping out the data associated with a table when we remove and then deploy again.

registry/AwsDynamoDB/src/index.js
```diff
@@ -1,345 +1,54 @@
...
-    )
-  },
+    async deploy(prevInstance, context) {
+      const tableName = get('tableName', this)
 
-  async getItem(prevInstance, context) {
-    if (!prevInstance.tables || prevInstance.tables.length === 0) {
-      return {}
+      context.log(`Creating table: '${tableName}'`)
+      await createTable(this)
```
brianneisler: @pmuens one thing to consider about the idempotency of create is that it's a bit different than the remove. When remove errors because the table is already gone, there's nothing else to do. However, in the create case, if create fails because the table already exist, it does not mean that the table has been configured with the desired configuration. It is entirely possible that a user deploys a table, they wipe out their state for some reason, they make a change to the dynamo db in serverless.yml and then they deploy again. In that case, this current code would skip over the create because it already exists, but the table does not match what they expect it to be.

pmuens:
@brianneisler Quick update: I've just implemented the `updateTable` functionality which makes it possible to update an existing table.

This supports all attributes except the `GlobalSecondaryIndexes` property (a log message will be printed out so that the user is aware of this).

Support for this is way more sophisticated and has to deal with enhanced state storage. AFAIK other components (e.g. `AwsS3Bucket`) are affected by this as well. I'm still looking into this but don't wanted to block this PR from being merged for the time-being...

registry/AwsDynamoDb/src/index.js
```diff
@@ -0,0 +1,75 @@
...
+        attributeDefinitions: this.attributeDefinitions,
+        provisionedThroughput: this.provisionedThroughput,
+        globalSecondaryIndexes: this.globalSecondaryIndexes,
+        sseSpecification: this.sseSpecification,
+        streamSpecification: this.streamSpecification
+      }
+      const prevInputs = prevInstance ? pick(keys(inputs), prevInstance) : {}
+      const configChanged = not(equals(inputs, prevInputs))
+
+      if (!prevInstance) {
```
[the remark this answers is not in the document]
pmuens: 😅 🤦‍♂️

pmuens:
@brianneisler thanks for the review! 👍

The idempotency logic is way more complex than I thought 😅

I just rebased and updated the PR. While doing so I added a new `ensureTable` function which takes in the desired function (such as `createTable`) and then tries to call this function. If it fails we try to recover by calling the corresponding "resuce" function (`updateTable` in this case).

Doing so we have one function where we pipe everything through while still using the "core" functions like `createTable` and `updateTable` behind the scenes, making the codebase easier to test and cleaner.

I added new tests as well to cover all those edge-cases.

Let's start the next round of reviews! 🥊 Ding, ding, ding 🥊

😹

pmuens:
> LGTM. Awesome work @pmuens. Love the ensureTable implementation.

Thanks @brianneisler 👍

Updating the failing tests...

pmuens:
@brianneisler jus updated the failing tests (replaced `loadType` with `import`).

PR #344, opened 2018-11-08
Add remove idempotency for AWS components

pmuens:
This PRs adds idempotency guarantees on removals for AWS components.

The implementation is based on the error codes AWS provides. This way we're not checking error-messages anymore which might be changed in the future. Unit tests were added to ensure the functionality.

**Note 1:** Right now the `AwsApiGateway` component was not updates since the API docs and its error codes are unavailable ATM --> https://docs.aws.amazon.com/goto/WebAPI/apigateway-2015-07-09/DeleteRestApi

**Note 2:** I've implemented this against the AWS SDK docs and checked the error structure (how to access the `code`) via `AwsIamRole` manually. Assume that not all components were tested manually and that the code I've picked from the docs might be wrong (updating this is a no-brainer though).

AWS components updated:

- [ ] `AwsApiGateway`
- [x] `AwsDynamoDB` (this was done in https://github.com/serverless/components/pull/341)
- [x] `AwsEventsRule`
- [x] `AwsIamPolicy`
- [x] `AwsIamRole`
- [x] `AwsLambdaFunction`
- [x] `AwsS3Bucket`
- [x] `AwsS3Website`
- [x] `AwsSnsSubscription`
- [x] `AwsSnsTopic`

/cc @brianneisler

### DIFF
--- a/__mocks__/aws-sdk.js
+++ b/__mocks__/aws-sdk.js
@@ -1,45 +1,97 @@
 const mocks = {
   // S3
   createBucketMock: jest.fn().mockReturnValue('bucket-abc'),
-  deleteBucketMock: jest.fn(),
+  deleteBucketMock: jest.fn().mockImplementation((params) => {
+    if (params.Bucket === 'already-removed-bucket') {
+      const error = new Error()
+      error.code = 'NoSuchBucket'
+      return Promise.reject(error)
+    }
+    return Promise.resolve()
+  }),
   listObjectsMock: jest.fn().mockImplementation((params) => {
-    if (params.Bucket === 'some-already-removed-bucket') {
-      return Promise.reject(new Error('The specified bucket does not exist'))
+    if (params.Bucket === 'already-removed-bucket') {
+      const error = new Error()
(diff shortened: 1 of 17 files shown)


pmuens:
Yes, good question @dschep 👍

All the SDK params are supported in this implementation so you should be able to configure it via `inputs` (if it's possible). Not sure if the `Retain` flag is CloudFormation specific.

brianneisler:
@dschep i'm not sure checking for existence is any faster from a performance perspective. Also you cannot guarantee that a deletion hasn't occurred between the time that you read and the time you request the removal, so either way we would still have to add the try/catch statements

plfx:
This branch has a build failure https://travis-ci.org/serverless/components/jobs/452373195

brianneisler:
Looks like an issue with the folder-hash module. Doesn't support node 6 in the engine declaration.
https://github.com/marc136/node-folder-hash/blob/master/package.json#L35
https://github.com/marc136/node-folder-hash/issues/12

dschep:
That's a good reason @brianneisler :+'

PR #346, opened 2018-11-12
Add AwsApiGateway remove idempotency

pmuens:
This is a follow-up PR of https://github.com/serverless/components/pull/344

After searching around I was able to find the following docs which include the error code used in this implementation: https://docs.aws.amazon.com/apigateway/api-reference/handling-errors/

### DIFF
--- a/registry/AwsApiGateway/src/index.js
+++ b/registry/AwsApiGateway/src/index.js
@@ -192,9 +192,9 @@ const AwsApiGateway = function(SuperClass) {
       try {
         context.log(`Removing API Gateway: "${this.apiName}"`)
         await deleteApi(APIGateway, { apiName: this.apiName, id: this.id })
-      } catch (e) {
-        if (!e.message.includes('Invalid REST API identifier specified')) {
-          throw e
+      } catch (error) {
+        if (error.code !== 'NotFoundException') {
+          throw error
         }
       }
     }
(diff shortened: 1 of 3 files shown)


registry/AwsApiGateway/src/index.js
```diff
@@ -192,9 +192,9 @@ const AwsApiGateway = function(SuperClass) {
       try {
         context.log(`Removing API Gateway: "${this.apiName}"`)
         await deleteApi(APIGateway, { apiName: this.apiName, id: this.id })
-      } catch (e) {
-        if (!e.message.includes('Invalid REST API identifier specified')) {
-          throw e
+      } catch (error) {
+        if (error.code !== 'NotFoundException') {
```
eahefnawy: wouldnt it be better if we made that check from prevInstance instead? otherwise if I try to remove without deploying and there's something wrong with the request (ie. validation error) it would still throw an error. Imo remove should return right away if there's no prevInstance

brianneisler: @eahefnawy I'm not sure the combination you're describing can happen. Removes only happen on records that we have in state. There is no previous instance to check for in a remove. If the remove is running, it means it found the record in state and is executing remove against it. The point here is that the state may not be correct. If we have a state record for a resource that has already been removed from a provider (possibly manually from the provider's dashboard) then we need to handle the error case idempotently.

PR #349, opened 2018-11-13
Fix info support for services without components

pmuens:
Fixes the bug where services without `components` defined couldn't run `info`.

/cc @ac360

### DIFF
--- a/registry/Service/src/index.js
+++ b/registry/Service/src/index.js
@@ -1,4 +1,9 @@
 import path from 'path'
+
+// TODO: move this into @serverless/utils ?!
+import isTypeConstruct from '../../../dist/utils/type/isTypeConstruct'
+import isComponent from '../../../dist/utils/component/isComponent'
+
 import {
   append,
   map,
@@ -32,6 +37,7 @@ const Service = async (SuperClass, superContext) => {
   return class extends SuperClass {
     async construct(inputs, context) {
       await super.construct(inputs, context)
+      // construct function instances
       this.functions = await map(
         async (func, alias) =>
           context.construct(Fn, {
@@ -41,6 +47,24 @@ const Service = async (SuperClass, superContext) => {
           }),
         or(this.functions, {})
       )
+      // construct component instances
+      this.components = await map(async (component, key) => {
+        if (isComponent(component)) {
+          return component
+        } else if (isTypeConstruct(component)) {
+          // eslint-disable-next-line no-shadow
+          const { type, inputs } = component
+          const Type = await context.loadType(type)
+          component = await context.construct(Type, inputs)
+          if (!isComponent(component)) {
+            throw new Error(
+              `The component "${key}" is not of type Component (it's of type ${component.name})`
+            )
+          }
+          return component
+        }
+        throw new Error(`The provided component definition ${JSON.stringify(component)} is invalid`)
+      }, or(this.components, {}))
     }
 
     async define() {
--- a/registry/Service/src/index.test.js
+++ b/registry/Service/src/index.test.js
@@ -27,48 +27,122 @@ afterAll(() => {
 })
 
 describe('Service', () => {
-  it('should default components and functions to empty objects', async () => {
-    const context = await createTestContext()
-    const Service = await context.loadType('Service')
-    const service = await context.construct(Service, {})
+  let context
+  let Service
 
-    expect(resolve(service.components)).toEqual({})
-    expect(resolve(service.functions)).toEqual({})
+  beforeEach(async () => {
+    context = await createTestContext()
+    Service = await context.loadType('Service')
   })
 
-  it('should construct function instances when construct is called', async () => {
-    const context = await createTestContext()
-    const Service = await context.loadType('Service')
-    const Function = await context.loadType('Function')
+  describe('#construct()', () => {
+    it('should default components and functions to empty objects', async () => {
+      const service = await context.construct(Service, {})
 
-    const service = await context.construct(Service, {
-      functions: {
-        hello: {
-          functionName: 'hello'
-        }
-      }
+      expect(resolve(service.components)).toEqual({})
+      expect(resolve(service.functions)).toEqual({})
     })
 
-    expect(service.functions.hello).toBeInstanceOf(Function.class)
-    expect(resolve(service.functions.hello.functionName)).toBe('hello')
+    it('should construct function instances', async () => {
+      const Function = await context.loadType('Function')
+
+      const service = await context.construct(Service, {
+        functions: {
+          hello: {
+            functionName: 'hello'
+          }
+        }
+      })
+
+      expect(service.functions.hello).toBeInstanceOf(Function.class)
+      expect(resolve(service.functions.hello.functionName)).toBe('hello')
+    })
+
+    it('should creaate component instances', async () => {
+      const Component = await context.loadType('Component')
+      const component1 = await context.construct(Component, {})
+
+      const service = await context.construct(Service, {
+        components: {
+          component1,
+          component2: {
+            type: './', // Service is of type Component
+            inputs: {}
+          }
+        }
+      })
+
+      expect(service.components.component1).toBeInstanceOf(Component.class)
+      expect(service.components.component2).toBeInstanceOf(Component.class)
+    })
+
+    it('should throw if type is not a component', async () => {
+      await expect(
+        context.construct(Service, {
+          components: {
+            component: {
+              type: '../Object', // Object is not of type Component
+              inputs: {}
+            }
+          }
+        })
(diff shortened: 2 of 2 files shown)


registry/Service/src/index.js
```diff
@@ -41,6 +43,14 @@ const Service = async (SuperClass, superContext) => {
           }),
         or(this.functions, {})
       )
+      // construct component instances
+      this.components = await map(
```
brianneisler: @pmuens why is this needed? Components should already be instances at this point since they are interpreted from the type/inputs combo in props.

registry/Service/src/index.test.js
```diff
@@ -53,6 +53,20 @@ describe('Service', () => {
...
   })
 
+  it('should construct component instances when construct is called', async () => {
+    const context = await createTestContext()
+    const Service = await context.loadType('Service')
+    const Component = await context.loadType('Component')
+
+    const service = await context.construct(Service, {
+      components: {
+        hello: {}
```
brianneisler: This is not a valid way of specifying components since it's not a component

pmuens:
Thanks for reviewing @dschep and @brianneisler 👍

> @pmuens I don't quite understand what the issue is but I know we shouldn't need to construct components like this. Can you elaborate a bit more on your understanding of the issue so we can figure out what a proper solution might be?

That is a good question @brianneisler... The problem right now is that if you create a new service (let's say it includes new `functions` via the `functions` section  but **NO** `components` are declared via the `components` section) inside an application like the SWAT you'll get an error when you deploy the application with this new service since the service tries to run `info` on each component (but there's none in our case).

I added a regression test which illustrates the behavior (https://github.com/serverless/components/blob/info-for-componentless-service/registry/Service/src/index.test.js#L89-L112). If you run this against the current "master" it'll fail. This PR attempts to fix this.

brianneisler:
@pmuens i think the actual issue might be here 

https://github.com/serverless/components/blob/57afef7e8265060d6ae761a1736f4055590738c4/registry/Component/src/index.js#L13

We're blindly assigning components either from inputs or from properties. What we should be doing is validating that the value we've been given for components is in fact an object or an array and that the values within that object or array are all component instances. If they are not, we should throw an error so that the user knows that they have a mistake in one of their components

pmuens:
@brianneisler thanks for looking into this 👍

I looked into this today, but it's not clear to me how to fix this. The issue is that `this.components` (in https://github.com/serverless/components/blob/57afef7e8265060d6ae761a1736f4055590738c4/registry/Component/src/index.js#L13) is the `serverless.yml` definition which can look something like this:

```yml
{ nestedService: { type: './nested-service' } }
```

My idea was to create the child components in the `Component` construct function like this:

```javascript

      // TODO BRN: this is not the best solution, this is causing problems
      this.instanceId = context.generateInstanceId()

      this.components = await map(
        async (comp) =>
          context.construct(Component, {
            ...comp
          }),
        or(get('components', inputs), this.components, {})
      )
}
```

This fails for various reasons. The main issue is that `Component` is invalid when I pass it into `super.construct`.

---

Aside form that here's a simple service setup which helps to reproduce the issue:

```yml
# ./serverless.yml

name: ComponentsTest
extends: Service

components:
  nestedService:
    type: ./nested-service
```

```yml
# ./nested-service/serverless.yml

name: NestedService
extends: Component
```

brianneisler:
@pmuens I think the misunderstanding here is how components are getting turned into instances vs how the functions are getting turned into instances.

In the usual case of components, they are declared in the serverless.yml like this...
```yaml
components:
  foo:
    type: FooComponent
    inputs: ...
```
When we interpret the yml, we convert any combination of `type` and `inputs` into an instance. This is similar to the idea of constructing an instance in code before passing it into a constructor as a parameter.

Now, moving to the `functions` of a service. These are not declared in yml using the `type` and `inputs` combination. So, they are not converted into instances when we interpret the yaml. Instead, the Service takes those plain objects and constructs `Function` instances from them in the `Service`'s `construct` method. 

Now, let's look at what you are trying to do in your tests.

In your tests, you are not using serverless.yml to declare your components. Instead, you are passing in your components using inputs programmatically. So there is no interpretation made against the components. The base level `Component` has no code in place to convert plain objects like this...
```js
{
  type: 'SomeComponent',
  inputs: ...
}
```
into an actual component. 

So, I think we can do a few things here (or pick and choose from the below list)

1) We add this as a feature to the low level `Component`, where it would scan the components during construct and convert any plain object `type` and `inputs` combinations into instances. The interpreter would still be responsible for converting any of these combinations found in yaml but this would give us support for when values like these are passed in programmatically via inputs.

2) We add some validation logic to the construct method of component. If it finds an instance or anything else that is not a Component on the `this.components` property, it throws an error to let the user know they have done something wrong (like passing in a plain object with `type` and `inputs` programmatically).

3) We update the tests accordingly to cover cases where a component instance is passed into a component... 
```js
 const service = await context.construct(Service, {
      functions: {
        hello: {
          functionName: 'hello'
        }
      },
      components: {
        myComponent: await context.construct(MyComponent, {
          some: 'input'
        })
      }
    })
```

where plain objects with `type` and `inputs` are present (this should either throw an error or create an instance depending on what we do above)...

```js
 const service = await context.construct(Service, {
      functions: {
        hello: {
          functionName: 'hello'
        }
      },
      components: {
        myComponent: {
          type: 'MyComponent',
          inputs: {
            some: 'input'
          }
       }
      }
    })
```

and, finally, where some meaningless object is provided in inputs (this should throw an error)

```js
 const service = await context.construct(Service, {
      functions: {
        hello: {
          functionName: 'hello'
        }
      },
      components: {
        myComponent: {
          foo: 'bar'
        }
      }
    })
```

brianneisler:
@pmuens also one other note. In your example
```js

  // TODO BRN: this is not the best solution, this is causing problems
  this.instanceId = context.generateInstanceId()

  this.components = await map(
    async (comp) =>
      context.construct(Component, {
        ...comp
      }),
    or(get('components', inputs), this.components, {})
  )
}
```

you'll need to consider the idea that you may encounter a mix. Some values you are iterating over may already be component instances, some may be garbage objects, and some may need conversion into components. In the final case, i don't think you want to use the base level `Component` to construct them (unless we can think of a use case for this). You'll instead need to use whatever type is indicated by the `type` property. 

You should be able to handle that doing something like this...
```js
if (isTypeConstruct(comp)) {
  const { type, inputs } = comp
  const Type = await context.loadType(type)
  const instance = context.construct(Type, inputs)
  if (!isComponent(instance)) {
    throw new Error('blast! this is not a component!')
  }
}
```

brianneisler:
@pmuens good change, but it looks like it broke a bunch of tests ;)

I've removed the snyk integration for now. We should be able to rely on the status checks again.

pmuens:
Hey @brianneisler thanks for the deep dive into this...

That makes sense. I just tested the changes you suggested locally and it seems to work. 

Should we move `isTypeConstruct` and `isComponent` to the `@serverless/utils` project? Right now they are in `src/utils` which means the components themselves don't have "direct access".

brianneisler:
@pmuens i think it makes sense to keep the isTypeConstruct and isComponent methods in core in case we ever need to change these methods. Instead, what we could do is export them at the top level of core. This will enable components to import these methods directly. For right now we can simply import within components by path, once we separate core and the components, we can make the core a peer dependency and then import from the node module

pmuens:
@brianneisler thanks again for the review and the in-depth explanation here.

I just rebased and updated the PR according to your suggestions. Added unit tests which cover all the  different cases as well.

Furthermore I tested it manually with the following service (which now successfully deploys):

```yml
# ./serverless.yml

name: TestService
extends: Service

credentials:
  type: EnvironmentBasedConfiguration
  inputs:
    variables:
      - AWS_ACCESS_KEY_ID
      - AWS_SECRET_ACCESS_KEY
      - AWS_SECURITY_TOKEN
      - AWS_SESSION_TOKEN

providers:
  aws:
    type: AwsProvider
    inputs:
      credentials:
        accessKeyId: ${this.credentials.values.AWS_ACCESS_KEY_ID}
        secretAccessKey: ${this.credentials.values.AWS_SECRET_ACCESS_KEY}
        securityToken: ${this.credentials.values.AWS_SECURITY_TOKEN}
        sessionToken: ${this.credentials.values.AWS_SESSION_TOKEN}
      region: us-east-1

components:
  nestedService:
    type: ./nested-service
    inputs: {}
  testComponent:
    type: AwsSnsTopic
    inputs:
      provider: ${this.providers.aws}
```

```yml
# ./nested-service/serverless.yml

name: NestedService
extends: Component
```

pmuens:
Thanks @brianneisler 👍

Pretty much just followed your recommendations 🙌

PR #352, opened 2018-11-13
Inline IamRole Policies & better default creation for lambdas

dschep:
## What has been implemented?

* AwsIamRole now supports being given an inline policy in addition to one with an ARN
* AwsIamRole no longer defaults to attaching the admin policy
* AwsLambdaCompute now specifies the minimum inline role that a lambda needs
  * doesn't currently set account Id in the ARN, not sure if that's available in this context?

## Steps to verify

* run `components deploy` in a service using AwsLambdaCompute without roles specified
* in the AWS console check that the roles created have an inline policy granting perms to create & write to a the right cloudwatch log group and not have the managed AdministratorAccess policy attached.

## Todos:

* [x] Write tests
* [x] Run Prettier

### DIFF
--- a/registry/AwsIamRole/src/index.js
+++ b/registry/AwsIamRole/src/index.js
@@ -122,11 +122,8 @@ const AwsIamRole = async (SuperClass, superContext) => {
         )
       }
 
-      // HACK BRN: Temporary workaround until we add property type/default support
-      const defaultPolicy = {}
       this.provider = inputs.provider
       this.service = inputs.service
-      this.policy = resolvable(() => or(inputs.policy, defaultPolicy))
       this.roleName = resolvable(() => or(inputs.roleName, `role-${this.instanceId}`))
     }
 
--- a/registry/AwsIamRole/src/index.js
+++ b/registry/AwsIamRole/src/index.js
@@ -1,19 +1,47 @@
-import { get, equals, is, resolve, sleep, or, resolvable, not, pick, keys } from '@serverless/utils'
+import {
+  get,
+  equals,
+  is,
+  isEmpty,
+  has,
+  resolve,
+  sleep,
+  or,
+  resolvable,
+  not,
+  pick,
+  keys
+} from '@serverless/utils'
 
-const attachRolePolicy = async (IAM, { roleName, policy }) => {
-  await IAM.attachRolePolicy({
-    RoleName: roleName,
-    PolicyArn: policy.arn
-  }).promise()
+const addRolePolicy = async (IAM, { roleName, policy }) => {
+  if (has('arn', policy)) {
+    await IAM.attachRolePolicy({
+      RoleName: roleName,
+      PolicyArn: policy.arn
+    }).promise()
+  } else if (!isEmpty(policy)) {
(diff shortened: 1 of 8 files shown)


registry/AwsIamRole/src/index.js
```diff
@@ -85,9 +113,7 @@ const AwsIamRole = async (SuperClass, superContext) => {
       await super.construct(inputs, context)
 
       // HACK BRN: Temporary workaround until we add property type/default support
-      const defaultPolicy = {
-        arn: 'arn:aws:iam::aws:policy/AdministratorAccess'
-      }
+      const defaultPolicy = {}
```
pmuens: Should we have a "sane" default here? In case it should we fail if no policy / policy arn is provided might want to move the validation into the `serverless.yml` via RAML.

registry/AwsLambdaFunction/src/index.js
```diff
@@ -150,7 +150,26 @@ const AwsLambdaFunction = async (SuperClass, superContext) => {
...
             service: 'lambda.amazonaws.com',
-            provider: this.provider
+            provider: this.provider,
+            policy: {
+              Version: '2012-10-17',
+              Statement: [
+                {
+                  Action: ['logs:CreateLogStream'],
+                  Resource: [
+                    `arn:aws:logs:us-east-1:*:log-group:/aws/lambda/${this.functionName}:*`
```
pmuens: We might want to extract the region from `this.provider` and string-interpolate it here.

Something like:

```suggestion
                    `arn:aws:logs:${this.provider.region}:*:log-group:/aws/lambda/${this.functionName}:*`
```

registry/AwsLambdaFunction/src/index.js
```diff
@@ -150,7 +150,26 @@ const AwsLambdaFunction = async (SuperClass, superContext) => {
...
+                  ],
+                  Effect: 'Allow'
+                },
+                {
+                  Action: ['logs:PutLogEvents'],
+                  Resource: [
+                    `arn:aws:logs:us-east-1:*:log-group:/aws/lambda/${this.functionName}:*:*`
```
pmuens: Same here...

```suggestion
                    `arn:aws:logs:${this.provider.region}:*:log-group:/aws/lambda/${this.functionName}:*:*`
```

registry/AwsIamRole/src/index.js
dschep: I couldn't think of a "sane" default. A Role without any policies is entirely possible, right? The question then is, is it ever something a user would want?

eahefnawy:
@dschep how about this API call?
https://github.com/serverless/serverless/blob/master/lib/plugins/aws/provider/awsProvider.js#L435

dschep:
Looks perfect @eahefnawy, Thanks!

registry/AwsIamRole/src/index.js
brianneisler: I think it makes sense to make this value required. If the user wants an empty policy than they can provide an empty object. Otherwise, it seems more likely that they simply forgot to assign a policy or the value was mistakenly undefined.

PR #354, opened 2018-11-15
Add AwsProvider config validation

pmuens:
This adds a basic validation to check whether the necessary AWS configuration is properly set.

/cc @shortjared

### DIFF
--- a/registry/AwsProvider/src/index.js
+++ b/registry/AwsProvider/src/index.js
@@ -1,15 +1,37 @@
-import { resolve } from '@serverless/utils'
+import { isEmpty } from '@serverless/utils'
 import AWS from 'aws-sdk'
 
-const AwsProvider = {
-  getSdk() {
-    // TODO BRN: This won't work for multi provider/region
-    AWS.config.update({ region: resolve(this.region), credentials: resolve(this.credentials) })
-    return AWS
-  },
-  getCredentials() {
-    return { region: this.region, credentials: this.credentials }
+const AwsProvider = (SuperClass) =>
+  class extends SuperClass {
+    async construct(inputs, context) {
+      await super.construct(inputs, context)
+
+      // NOTE: we cannot use `resolvable` or `resolve` here since AwsProvider doesn't extend Component
+      this.region = inputs.region || 'us-east-1'
+      this.credentials = inputs.credentials
+    }
+
+    getSdk() {
+      this.validate()
+      // TODO BRN: This won't work for multi provider/region
+      AWS.config.update({ region: this.region, credentials: this.credentials })
+      return AWS
+    }
+
+    getCredentials() {
+      this.validate()
+      return { region: this.region, credentials: this.credentials }
+    }
+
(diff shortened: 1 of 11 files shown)


brianneisler:
@pmuens good change but it looks like a bunch of tests are failing now ;)

I've disabled the snyk integration so we should be able to lean on status checks again.

registry/AwsProvider/src/index.js
```diff
@@ -1,14 +1,33 @@
-import { resolve } from '@serverless/utils'
+import { resolve, isEmpty } from '@serverless/utils'
 import AWS from 'aws-sdk'
 
 const AwsProvider = {
```
brianneisler: Perhaps we should default the region to `us-east-1` in the construct method.

pmuens:
D'oh. Sorry for the sloppiness on my end @brianneisler 

Only ran the `AwsProvider` tests locally. 🤦‍♂️ 

Just reabsed and updated the PR so that a default region of `us-east-1` is set. Furthermore I fixed all the other component tests where the `AwsProvider` is used.

registry/AwsProvider/src/index.js
dschep: Probably the best is `inputs.region || process.env.AWS_DEFAULT_REGION || 'us-east-1'`

pmuens:
Just rebased and updated this PR so that it doesn't conflict anymore.

registry/AwsProvider/src/index.js
```diff
@@ -1,15 +1,37 @@
...
+      return AWS
+    }
+
+    getCredentials() {
+      this.validate()
+      return { region: this.region, credentials: this.credentials }
+    }
+
+    validate() {
+      if (!/.+-.+.\d+/.test(this.region)) {
```
eahefnawy: hmm I think it'd be more accurate if we validate against the AWS list of regions. Something like

```
contains(awsRegions, this.region)
```

dschep: Easy way to get them all:
```
aws ec2 describe-regions
```

If we do this, I think we should have a test that checks that the list is upto date so that we get errors if AWS adds a new region. JS equivalent of the above CLI call:
```javascript
import AWS from 'aws-sdk'
const ec2 = new AWS.EC2()
const regions = await ec2.describeRegions().promise()
```

brianneisler: Minor issue here. Hard coding the regions prevents earlier versions from being able to use new regions when they are released. 

I like the idea of checking against the actual region list, however, we ran into issues here with this validation check being synchronous. Getting the region list is an async process which poses a problem.

I think what Philip has here is good enough for now and we can improve this once we introduce an async validation process.

pmuens:
Thanks for the reviews @brianneisler @eahefnawy and @dschep 👍

Merging since we might add a check against the actual regions later on (see: 
https://github.com/serverless/components/pull/354#discussion_r235394437)

PR #360, opened 2018-11-23
Add "or" / || support for Serverless Variables

pmuens:
This PR adds the first pass of "or" / `||` support in Serverless Variables.

Tests with some edge-case scenarios were added, however I'm not sure how far we want to get with this or implementation here (e.g. support for nesting).

---

Steps to validate it:

```yml
# ./serverless.yml

name: ComponentsTest
extends: Service

components:
  nestedService:
    type: ./nested-service
    inputs:
      name: ${this.name || 'SomeDefaultName'}
      description: ${this.notSet || false || undefined || this.extends}
```

```yml
# ./nested-component/serverless.yml

name: NestedService
extends: Component

inputTypes:
  name:
    type: string
    required: true
  description:
    type: string
```

```javascript
const { resolve } = require('@serverless/utils')

const NestedService = (SuperClass) =>
  class extends SuperClass {

      console.log('---')
      console.log(`Name: ${resolve(inputs.name)}`)
      console.log(`Description: ${resolve(inputs.description)}`)
      console.log('---')
    }

    shouldDeploy(prevInstance) {
      return 'deploy'
    }

    deploy(prevInstance, context) {
      context.log('Deploying...')
    }
  }

module.exports = NestedService
```

Run `components deploy` and see the output on the cosole


src/utils/variable/regexVariable.js
```diff
@@ -1,4 +1,4 @@
-const def = '\\${([a-zA-Z0-9_\\.[\\]]*)}'
+const def = '\\${(([a-zA-Z0-9_\\.[\\]]*)[\\s\\|\\|\\s]*([a-zA-Z0-9_\\.\\\'\\"]*))}'
```
dschep: Since this is already just a string (as opposed to a regex literal `/foobar/`), I think you should split this up and add comments.
```suggestion
const def = '\\${' + // opening ${
  '(([a-zA-Z0-9_\\.[\\]]*)' + // match number of any alpha numeric, ., [, and ]
  '[\\s\\|\\|\\s]*' + // match ' || ' 
  '([a-zA-Z0-9_\\.\\\'\\"]*))' +  // match number of any alpha numeric, ., ", and '
  '}' // closing }
```

In fact.. breaking it up and adding comments forced me to review it... and i'm not sure about `[\\s\\|\\|\\s]*`. The comment I added is what I _think_ you wanted, but I think the regex for that would be `(\s\\|\\|\s)`, where as what you have would match any combination of whitespace and `|`. IE, it would match ` || ` as well as `| |     | ||||   || ` which I don't think is right.

(I wish JS just had [python-style verbose regexes](https://docs.python.org/2/library/re.html#re.VERBOSE))

src/utils/variable/resolveVariableString.test.js
```diff
@@ -34,4 +34,15 @@ describe('#resolveVariableString()', () => {
       })
     ).toBe('dude, sweet')
   })
+
+  it('should be able to deal with OR (||) usages in variable resolutions', () => {
+    expect(resolveVariableString("Hello ${abc || 'world'}", { abc: false })).toBe('Hello world')
```
dschep: I know that this is the behavior of `||`, but is that actually the behavior we want in our config? This would make it impossible to define a boolean value that's imported from elsewhere but defaults to true (as in, it gets set to true when the imported value is false, which seems wrong). This would, imo, be cause for sticking with a `,` delimiter like we do in V1.

src/utils/variable/regexVariable.js
dschep: infact maybe the regex for ` || ` should be `(\s+\\|\\|\s+)` (the plus makes the whitespace optional, matching javascript semantics.

pmuens: Thanks for the in-depth review on this on @dschep 👍

src/utils/variable/resolveVariableString.test.js
pmuens: Yes, good question. /cc @brianneisler

dschep: Was thinking a bit more about this on my morning bike ride today... Another reason I prefer `,` over js-style ` || `: Users would probably expect you to be able to fallback to something other than a string literal(`${this.foo || this.bar}`) in addition to expecting to be able to use it more than once(`${this.foo || this.bar || 'foobar'}`).

brianneisler: @dschep i'm not quite following your line of thinking. The logical OR syntax in JS returns the first truthy value. This means you should be able to cover all the cases you've mentioned that the comma syntax would cover.  Also, agree that this should work with multiple OR and any value. I sent @pmuens a private message discussing some things to cover but i'll add it as a comment to this PR to ensure it's public.

brianneisler:
@pmuens, I just reviewed the PR for the logical OR syntax in the variables. Your solution addresses the immediate need but I don’t think it takes into account the longer term picture. Variables are as complicated as a language itself and we need to lay the foundation for handling the problem. I’d like you to take a step back from the problem and put your architect hat on.

Variables are conceptually the same as an expression in a language. What are the steps usually involved in converting a string expression (code) into something that’s executable?

Some other things to consider
- this only handles one OR, what if there is more than one?
- what if two or more separate variables are OR’d together?
- you have strings embedded into the OR syntax. This is good, but what about other types? Number, Boolean, etc.
- when more than one variable reference is used, we will need a way to pull out all of the reference ids from each variable.

I think before more code is implemented for this it would be good to write up a simple one pager on a design to tackle this problem for the long term and outline what we can do immediately that is inline with the long term architecture

src/utils/variable/resolveVariableString.test.js
dschep: What I'm suggesting is that logical OR isn't necessarily what the variable system should be supporting. I'd think what we're looking to implement is a system for variables to have fallback values. Logical OR means that you can't have a fallback value for a variable that can be false. I think an operator that does the same as `variable===undefined?'fallback':variable` is more desirable than logical OR (`variable?variable:'falback'` in ternary form to compare to the other)

brianneisler: @dschep ahh i see your point. That said, what I imagine for the next generation of variables is much more in line with terraforms interpolated syntax. Which is to say, very scriptish. This gets us the flexibility of a simple scripting language right within the `serverless.yml` and opens up a lot of doors without the user having to go and add in external js files (not that we shouldn't support that as well).

@pmuens one thing that might get us much farther much faster is to consider using an existing javascript AST parser that allows you to go from code -> AST 

something like acorn or recast
https://github.com/acornjs/acorn
https://github.com/benjamn/recast

using this, you could convert to an AST. We could analyze the AST for property references and pull out the reference ids that we need. We could also detect code concepts that we don’t want to support within variables (like eval, classes, declaring new functions, etc)

once that’s done, interpreting the variable is just a matter of  `new Function()` and you can just take the data object and inject each property by name as a parameter, which gets you what you need for `this`, `context`, `inputs`, etc. You can also inject any number of utility methods for use by the user (a number of which can come from the utils project)

brianneisler:
> Tests for something like ${this.foo || this.bar} might be helpful since IDK if this works right now
@pmuens +1 on this comment

src/utils/variable/regexVariable.js
```diff
@@ -1,4 +1,4 @@
-const def = '\\${([a-zA-Z0-9_\\.[\\]]*)}'
+const def = "\\${([a-zA-Z0-9_\\.[\\]\\|\\s\\']*)}"
```
dschep: It's not quite as complicated as before, but it'd still be nice for this to be spread over a few lines and commented.

src/utils/variable/newVariable.js
```diff
@@ -9,20 +18,33 @@ const newVariable = (variableString, data) => ({
...
-    return walkReducePath(
-      (instanceIds, value) => {
-        if (isVariable(value)) {
-          return concat(instanceIds, value.findInstanceIds())
-        }
-        if (has('instanceId', value)) {
-          return append(get('instanceId', value), instanceIds)
-        }
-        return instanceIds
+    const variableStrings = variableString
```
brianneisler: Using an AST parser will be necessary here to collect all of the property paths from any expression.

src/utils/variable/regexVariable.js
dschep: Will that cause issues if some one tries to use `{` or `{` inside the `${}`? If so, then it can be: `'\\${([^{}]*)}'`

src/utils/variable/resolveVariableString.js
```diff
@@ -1,16 +1,36 @@
...
 
 const resolveVariableString = (variableString, data) => {
-  const { exact, expression, match } = matchVariable(variableString)
+  const { exact, match, expression } = matchVariable(variableString)
+
   if (!match) {
     return variableString
   }
+
   const resolvedExpression = resolveVariableString(expression, data)
```
brianneisler: The `resolvedExpression` is actually the function body. This recursive pass allows for variables embedded within variables to be resolved and then then finalized expression to be interpreted as a variable.

src/utils/variable/resolveVariableString.js
```diff
@@ -1,16 +1,36 @@
...
   let value = get(resolvedExpression, data)
```
brianneisler: Don't need the get anymore. Instead we should just always use the function method for resolving variables.

src/utils/variable/resolveVariableString.js
```diff
@@ -1,16 +1,36 @@
...
+      self = args[thisPos]
+      // remove `this` value(s) from the params and args array
+      params.splice(thisPos, 1)
+      args.splice(thisPos, 1)
+    }
+
+    const func = new Function(params, `return ${funcBody}`)
+    const res = func.apply(self, args)
+
+    value = variableString.replace(match, res)
```
brianneisler: This results in the value always being a string. We can fix this by only doing this string replacement when it's not an exact match.

src/utils/variable/regexVariable.js
brianneisler: @dschep there are few cases where that makes sense, such as declaring an object within a variable.

`${ { foo: this.foo } }`

We also support nested variables

`${foo${bar}}`

PR #362, opened 2018-11-26
Add support for variable child component definitions with any depth

pmuens:
This PR adds support for child component definitions at various places in the component instance (rather than only `components`).

Here's how one can validate that it works (or take a look at the tests which ships with this PR):

```yml
# ./serverless.yml

name: ComponentsTest
extends: Service

components:
  nestedService:
    type: ./nested-service
```

```yml
# ./nested-service

name: NestedService
extends: Component
```

```javascript
// ./nested-service/index.js
const { resolve } = require('@serverless/utils')

const NestedService = {
  define() {
    return {
      components: {
        foo: {
          type: 'AwsS3Bucket',
          inputs: {}
        }
      },
      fn: {
        type: 'AwsLambdaFunction',
        inputs: {}
      }
    }
  }
}

module.exports = NestedService
```

/cc @brianneisler

### DIFF
--- a/registry/Component/src/index.js
+++ b/registry/Component/src/index.js
@@ -1,5 +1,9 @@
-import { append, get, or, pick, reduce, equals } from '@serverless/utils'
+import { omit, append, set, get, or, pick, equals, walkReduce } from '@serverless/utils'
 import { pickComponentProps } from './utils'
+
+// TODO: move this into @serverless/utils ?!
+import isTypeConstruct from '../../../dist/utils/type/isTypeConstruct'
+import { isComponent, walkReduceComponentChildren } from '../../../dist/utils/component'
 
 const DEPLOY = 'deploy'
 
@@ -29,9 +33,17 @@ const Component = (SuperClass) =>
     }
 
     async define() {
-      return {
-        ...or(this.components, {})
-      }
+      return walkReduce(
+        (accum, value, pathParts) => {
+          if (isTypeConstruct(value) || isComponent(value)) {
+            return set(pathParts, value, accum)
+          }
+          return accum
+        },
+        {},
+        // `parent` is omitted since this value is already resolved and considering it causes infinite loops
+        omit(['parent'], this)
+      )
     }
 
     shouldDeploy(prevInstance) {
@@ -52,10 +64,16 @@ const Component = (SuperClass) =>
     async remove() {}
 
     async info() {
-      const children = await reduce(
-        async (accum, component) => append(await component.info(), accum),
+      const children = await walkReduceComponentChildren(
+        async (accum, value) => {
+          if (isComponent(value)) {
+            const component = value
+            return append(await component.info(), accum)
+          }
+          return accum
+        },
         [],
-        or(this.components, {})
+        omit(['parent'], this)
       )
       return {
         title: this.name,
--- a/src/utils/component/defineComponent.js
+++ b/src/utils/component/defineComponent.js
@@ -1,16 +1,7 @@
-import {
-  filter,
-  forEach,
-  get,
-  isFunction,
-  isObject,
-  map,
-  or,
-  resolve,
-  clone
-} from '@serverless/utils'
+import { get, isFunction, or, resolve, clone, set, walkReduce } from '@serverless/utils'
 // import appendKey from './appendKey'
 // import getKey from './getKey'
+import isTypeConstruct from '../type/isTypeConstruct'
 import hydrateComponent from './hydrateComponent'
 import isComponent from './isComponent'
 import { SYMBOL_STATE } from '../constants'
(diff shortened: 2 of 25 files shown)


pmuens:
Just rebased and updated the PR.

Furthermore added some more tests. Should be good for an initial review.

registry/Component/src/index.js
```diff
@@ -1,6 +1,19 @@
...
+  or,
+  pick,
+  reduce,
+  equals,
+  walkReduceDepthFirst
+} from '@serverless/utils'
 import { pickComponentProps } from './utils'
 
+// TODO: move this into @serverless/utils ?!
+import isTypeConstruct from '../../../dist/utils/type/isTypeConstruct'
```
dschep: regardless of moving it or not... we shouldn't be importing from `dist`, right?

pmuens: Good thinking! The issue here has something todo with our project structure and Babel usage.

The problem is that we have one Babelified codebaase in `src`. This is transpiled into `dist`. And then we have another Baabelified codebase in `registry` which transpiles all the component files and puts them into `<component-name>/dist`. So if we want to access stuff from the "root" project (the one where the files are located in `src`) we need to refer to the transpiled files since old Node.js versions cannot understand the new shiny JavaScript syntax.

pmuens:
@dschep thanks for the review 👍

@brianneisler I just updated the code and included the support to resolve child components at any depth. So in conclusion you can now define child components at any prop-level in `serverless.yml` and those type definitions are then resolved and attached as child components.

Furthermore I added tests which validate that it works and cover the core functionalities. If you'd like to you can use the example in the PR description to give this functionality it a spin.

registry/Component/src/index.js
pmuens: Good questions. AFAIK the long-term plan is to have the registry (published components) live somewhere else like in an S3 bucket (we already have that functionality in place) and pull out all the core logic into the `@serverless/utils` package so that once the components are downloaded and installed the utils package is installed which includes all the logic. This makes versioning / releasing easier since utils is independent of the core and its releases.

@brianneisler migght know more here.

registry/Component/src/index.js
```diff
@@ -29,7 +42,23 @@ const Component = (SuperClass) =>
     }
 
     async define() {
+      // `components` are omitted because they are re-included in the return statement
+      // `parent` is omitted since this value is already resolved and considering it causes infinite loops
+      const filteredInstance = omit(['components', 'parent'], this)
+
+      const componentDefinitions = walkReduceDepthFirst(
+        (accum, value, pathParts) => {
+          if (isTypeConstruct(value)) {
```
brianneisler: Looking for type constructs might make sense, but we also need to find actual components. You'll want to also check for `isComponent` to find component instances that have already been constructed.

registry/Component/src/index.js
```diff
@@ -29,7 +42,23 @@ const Component = (SuperClass) =>
     }
 
     async define() {
+      // `components` are omitted because they are re-included in the return statement
+      // `parent` is omitted since this value is already resolved and considering it causes infinite loops
+      const filteredInstance = omit(['components', 'parent'], this)
```
brianneisler: `parent` makes sense to omit, i think a simpler solution here might just be to include `components` in the walk and use that to discover any components contained within the components property.

src/utils/component/defineComponent.js
```diff
@@ -41,9 +43,25 @@ const defineComponent = async (component, state, context) => {
   }
 
   component = hydrateComponent(component, state, context)
+
   if (isFunction(component.define)) {
     let children = await or(component.define(context), {})
-    children = filter(isComponent, map(resolve, children))
+    children = await walkReduceDepthFirst(
```
brianneisler: @pmuens one thought here is that you can use the `interpretProps` method to handle the portion responsible for converting type constructs into instances and interpreting variable strings. Once that pass has been made, you can then simply scan for components and filter out everything else.

registry/Component/src/index.js
pmuens: This has been fixed...

registry/Component/src/index.js
pmuens: This has been fixed...

registry/Component/src/index.js
```diff
@@ -29,7 +42,23 @@ const Component = (SuperClass) =>
...
+
+      const componentDefinitions = walkReduceDepthFirst(
```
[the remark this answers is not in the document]
pmuens: I see that and I tried to fix it but apparently it's not possible to omit components of components.

The reason being that once we encounter a component we assign it (and it's properties which can be components as well) to the object which is used as an accumulator in the `reduce`.

That way every property of this component (which can be a component itself) is copied over as well.

Here's a test case I tried to use to test for that:

```javascript
// registry/Component/src/index.test.js

it('should not consider components nested inside of other components', async () => {
  const component = await context.construct(Component, {})
  const instance = await context.construct(Component, {})

  component.component1 = {
    type: './',
    inputs: {},
    instance
  }

  component.component2 = {
    type: './',
    inputs: {},
    component3: {
      type: './',
      inputs: {}
    },
    instance
  }

  const children = await component.define()

  expect(children.component1).toEqual(component.component1)
  expect(children.component1.instance).toBeUndefined()
  expect(children.component2).toEqual(component.component2)
  expect(children.component2.instance).toBeUndefined()
  expect(children.component2.component3).toBeUndefined()
})
```

I implemented the filtering of components inside of components in the `defineComponent` function since that's the only place where logic like that can be implemented.

pmuens:
Closing this PR because of an upcoming update...

PR #363, opened 2018-11-27
Adds support for various returned objects in define method

eahefnawy:
In order to make it easier to programmatically define components we should make this function more like defining components in serverless.yml.
```
define() {
  return {
    foo: {
      type: ‘FooComponent’,
      inputs: ...
    },
    bar: {
      type: ‘./path/to/BarComponent’,
      inputs: ...
    }
  }
}
```
We should also support pointing to types that are already loaded. 

const FooComponent = await context.loadType(‘FooComponent’)
```
define() {
  return {
    foo: {
      type: FooComponent,
      inputs: ...
    }
  }
}
```
As well as component instances that have already been constructed

const FooComponent = await context.loadType(‘FooComponent’)
```
define() {
  return {
    foo: context.construct(FooComponent, {})
  }
}
```

### DIFF
--- a/registry/AwsProvider/src/index.js
+++ b/registry/AwsProvider/src/index.js
@@ -1,4 +1,3 @@
-import { isEmpty } from '@serverless/utils'
 import AWS from 'aws-sdk'
 
 const AwsProvider = (SuperClass) =>
@@ -26,10 +25,6 @@ const AwsProvider = (SuperClass) =>
     validate() {
       if (!/.+-.+.\d+/.test(this.region)) {
         throw new Error(`Invalid region "${this.region}" in your AWS provider setup`)
-      }
-
-      if (!this.credentials || isEmpty(this.credentials)) {
-        throw new Error(`Credentials not set in your AWS provider setup`)
       }
     }
   }
--- a/src/utils/component/defineComponent.js
+++ b/src/utils/component/defineComponent.js
@@ -3,11 +3,9 @@ import { filter, forEach, get, isFunction, isObject, map, or, resolve } from '@s
 // import getKey from './getKey'
 import hydrateComponent from './hydrateComponent'
 import isComponent from './isComponent'
+import isTypeConstruct from '../type/isTypeConstruct'
 // import setKey from './setKey'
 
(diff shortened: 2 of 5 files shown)


registry/AwsProvider/src/index.js
```diff
@@ -27,10 +26,6 @@ const AwsProvider = (SuperClass) =>
       if (!/.+-.+.\d+/.test(this.region)) {
         throw new Error(`Invalid region "${this.region}" in your AWS provider setup`)
       }
-
-      if (!this.credentials || isEmpty(this.credentials)) {
```
pmuens: This should be kept and extended to support environment based config rather than removed.

Furthermore it should be part of a different PR.

brianneisler: Agreed this should be part of a different PR

src/utils/component/defineComponent.js
```diff
@@ -18,6 +16,29 @@ const defineComponent = async (component, state, context) => {
...
+    //     context,
+    //     root: context.root,
+    //     path: context.root
+    //   },
+    //   context
+    // )
+    let typeQueryChildren = filter(isTypeConstruct, map(resolve, children))
+
+    typeQueryChildren = await map(async (typeQueryChild) => {
+      const child = await context.loadType(typeQueryChild.type)
```
pmuens: Just stumbled upon this one today... This needs to be renamed to `context.import`

pmuens:
This PR is related to #362

brianneisler:
@eahefnawy i believe @pmuens PR has covered what this PR originally intended to provide. We can probably close this PR in favor of https://github.com/serverless/components/pull/362

@pmuens one note is that we should also copy over the change to the `loadType` method from this PR that simply returns the type if the query is already a type.

eahefnawy:
closing, but not deleting branch. I'll leave it for @pmuens to copy any needed code from here.

pmuens:
FYI: @brianneisler @eahefnawy just ported the changes which were made to `loadType` over into #362 and added a unit tests as well.

PR #368, opened 2018-11-28
Supports Idempotency for AwsIamRole create method

eahefnawy:


registry/AwsIamRole/src/index.js
```diff
@@ -132,17 +132,81 @@ const AwsIamRole = async (SuperClass, superContext) => {
       this.arn = get('arn', prevInstance)
     }
 
-    shouldDeploy(prevInstance) {
-      const inputs = {
+    async shouldDeploy(prevInstance) {
+      // 1. run sls deploy for the first time: Framework creates role
```
pmuens: We should extract this into a separate doc so other component authors can read through this and follow along.

PR #370, opened 2018-12-03
Added sync method for AwsLambdaFunction

eahefnawy:


registry/AwsLambdaFunction/src/index.js
```diff
@@ -142,6 +143,34 @@ const AwsLambdaFunction = async (SuperClass, superContext) => {
...
+        this.role = {
+          arn: res.Role
+        }
+        this.handler = res.Handler
+        this.functionDescription = res.Description
+        this.timeout = res.Timeout
+        this.memorySize = res.MemorySize
+        this.environment = res.Environment ? res.Environment.Variables : {}
+      } catch (e) {
+        if (e.code === 'ResourceNotFoundException') {
```
pmuens: 👌

registry/AwsLambdaFunction/src/index.test.js
```diff
@@ -507,6 +507,67 @@ describe('AwsLambdaFunction', () => {
...
+
+    awsLambdaFunction = resolveComponentEvaluables(awsLambdaFunction)
+
+    const status = await awsLambdaFunction.sync()
+
+    expect(status).toBe('removed')
+  })
+
+  it('sync should update lambda config if lambda config changed in provider', async () => {
+    let awsLambdaFunction = await context.construct(AwsLambdaFunction, {
```
pmuens: It might be a good idea to update the `getFunctionConfiguration` mock funciton in `__mocks__/aws-sdk.js` so that it returns a specific object when a Lambda function with a prescriptive name is passed in (like we e.g. do with `already-removed-function`).

Looking at the test in isolation I had no idea why this would trigger the `sync` function so spit out a command to update the config.

We could use a function name like `to-update-function` or `function-to-update` to indicate that we need to update the lambda config.

eahefnawy: hmmm but in that case the mock would only support two "prospective names", one for removal and the other for deploy. Using the mock again we'd have to support more cases. I've purposefully kept it this way to make sure it always resolves OR rejects in a single case.

registry/AwsLambdaFunction/src/index.js
```diff
@@ -142,6 +143,34 @@ const AwsLambdaFunction = async (SuperClass, superContext) => {
...
+        this.functionName = res.FunctionName
+        this.runtime = res.Runtime
+        this.role = {
+          arn: res.Role
+        }
+        this.handler = res.Handler
+        this.functionDescription = res.Description
+        this.timeout = res.Timeout
+        this.memorySize = res.MemorySize
+        this.environment = res.Environment ? res.Environment.Variables : {}
```
dschep: Would it not also make sense to set `this.hash` and change the hash computation code locally to hash the packed zip so that you can return `undefined` from `shouldDepoy` when possible?

__mocks__/aws-sdk.js
```diff
@@ -84,6 +84,34 @@ const mocks = {
...
   }),
+  getFunctionConfigurationMock: jest.fn().mockImplementation((params) => {
+    if (params.FunctionName === 'already-removed-function') {
+      const error = new Error()
+      error.code = 'ResourceNotFoundException'
+      return Promise.reject(error)
+    }
+    const res = {
+      FunctionName: params.FunctionName,
+      FunctionArn: `rn:aws:lambda:us-east-1:xxx:function:${params.FunctionName}`,
```
pmuens: `arn`

registry/AwsLambdaFunction/src/index.js
eahefnawy: thanks! done that!

registry/AwsLambdaFunction/src/index.js
```diff
@@ -116,8 +104,20 @@ const AwsLambdaFunction = async (SuperClass, superContext) => {
       this.zip = get('zip', prevInstance)
     }
 
-    async shouldDeploy(prevInstance) {
-      this.hash = await hashCode(this.code)
+    async shouldDeploy(prevInstance, context) {
+      if (isArchivePath(this.code)) {
+        this.zip = readFileSync(this.code)
```
pmuens: `await readFile()`

examples/demo/SimpleService/serverless.yml
```diff
@@ -14,17 +14,28 @@ compute:
     runtime: nodejs
 
 functions:
-  IdempotentCron10:
+  decemberdemo:
```
pmuens: Does this example has anything todo with the PR?

We're constantly committing changes to the `SimpleService` which seems to be related to local testing and not specific to the PR itself.

package-lock.json
```diff
@@ -5,9 +5,9 @@
   "requires": true,
   "dependencies": {
     "@babel/cli": {
-      "version": "7.1.5",
-      "resolved": "https://registry.npmjs.org/@babel/cli/-/cli-7.1.5.tgz",
-      "integrity": "sha512-zbO/DtTnaDappBflIU3zYEgATLToRDmW5uN/EGH1GXaes7ydfjqmAoK++xmJIA+8HfDw7UyPZNdM8fhGhfmMhw==",
+      "version": "7.2.0",
```
pmuens: Can we please not include `package-lock.json` file changes in every PR?

PRs are super bloated and hard to digest when 90% of file changes are related to superflous lockfile changes.

examples/demo/SimpleService/serverless.yml
eahefnawy: yep, you're right.

registry/AwsLambdaFunction/src/index.js
```diff
@@ -1,7 +1,7 @@
 import path from 'path'
+import crypto from 'crypto'
 import { tmpdir } from 'os'
-import { readFileSync } from 'fs'
-import { hashElement } from 'folder-hash'
```
pmuens: `folder-hash` is still in the dependencies

registry/AwsLambdaFunction/src/index.js
```diff
@@ -1,7 +1,7 @@
...
+import { readFile } from 'fs-extra'
```
pmuens: This was not added to the dependencies, so the component is not self-contained.

You could also pull that from `@serverless/utils` --> https://github.com/serverless/utils/blob/master/src/fs/readFile.js

registry/AwsLambdaFunction/src/index.test.js
```diff
@@ -137,8 +137,6 @@ describe('AwsLambdaFunction', () => {
 
     awsLambdaFunction = resolveComponentEvaluables(awsLambdaFunction)
 
-    awsLambdaFunction.pack = jest.fn()
```
pmuens: Don't know where else I should put it, but the `folder-hash` module is still mocked at the top.

registry/AwsLambdaFunction/src/index.js
eahefnawy: @pmuens the utils version wont work because it parses the file, expecting it to be test/json/yml.

pmuens:
Approving so that we can move on...

PR #372, opened 2018-12-06
Adds support for idempotency for AwsS3Bucket deployment

eahefnawy:
## Steps to Verify

- make first deployment of new s3 bucket
- delete `.serverless` folder
- run deploy again -> framework will not deploy as no changes happened
- delete bucket from s3 console
- run deploy again -> framework will create bucket again even though it exists in `.serverless` folder

### DIFF
--- a/tests/integration/library-usage.test.js
+++ b/tests/integration/library-usage.test.js
@@ -28,7 +28,7 @@ describe('Integration Test - Library Usage', () => {
     })
   })
 
-  describe.only('when using a path to a project', () => {
+  describe('when using a path to a project', () => {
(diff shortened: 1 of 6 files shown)


tests/integration/library-usage.test.js
```diff
@@ -28,8 +28,10 @@ describe('Integration Test - Library Usage', () => {
     })
   })
 
-  describe('when using a path to a project', () => {
+  describe.only('when using a path to a project', () => {
```
pmuens: This should be removed...

eahefnawy: oh!!

PR #378, opened 2018-12-10
Omit internal types during status logging

pmuens:
This omits our internal types from being logged (such as `AwsProvider`) since those a quite verbose when logging and cause the `JSON.stringify` command to fail due to deeply nested structures.

### DIFF
--- a/src/utils/dag/logStatus.js
+++ b/src/utils/dag/logStatus.js
@@ -1,16 +1,25 @@
-import { filter, mapObjIndexed, resolve, not, isEmpty } from '@serverless/utils'
+import { filter, assoc, reduce, resolve, not, isEmpty } from '@serverless/utils'
 
 function getParameters(instance) {
   if (instance.inputs && instance.inputTypes) {
     const requiredParams = filter((inputType) => !!inputType.required, instance.inputTypes)
-    const paramsAsObject = mapObjIndexed(
-      (num, key) => resolve(instance.inputs[key]),
+    const paramsAsObject = reduce(
+      (accum, _, key) => {
+        const value = resolve(instance.inputs[key])
+        // TODO: replace with a universal util function which checks whether we're
+        // dealing with a Component, Object, Provider, etc. instance here
+        // omitting own Types due to circular structure / verbosity here
+        if (!(value.name && value.version)) {
+          return assoc(key, value, accum)
+        }
+        return accum
+      },
+      {},
       requiredParams
     )
+
     if (not(isEmpty(paramsAsObject))) {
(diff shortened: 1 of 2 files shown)


src/utils/dag/logStatus.js
```diff
@@ -1,16 +1,25 @@
...
     const requiredParams = filter((inputType) => !!inputType.required, instance.inputTypes)
-    const paramsAsObject = mapObjIndexed(
-      (num, key) => resolve(instance.inputs[key]),
+    const paramsAsObject = reduce(
+      (accum, _, key) => {
+        const value = resolve(instance.inputs[key])
+        // TODO: replace with a universal util function which checks whether we're
+        // dealing with a Component, Object, Provider, etc. instance here
+        // omitting own Types due to circular structure / verbosity here
+        if (!(value.name && value.version)) {
```
eahefnawy: I think it would be more accurate to use the `isType` util method here instead

src/utils/dag/logStatus.test.js
```diff
@@ -99,4 +100,36 @@ describe('#logStatus()', () => {
...
+            type: 'string',
+            required: true
+          }
+        }
+      }
+    }
+    logStatus(iteratee, node, context)
+
+    expect(mockLog).toHaveBeenCalledTimes(1)
+    expect(mockLog.mock.calls[0][0]).toMatch(/Deploying.*with parameters.*/)
```
eahefnawy: how does this assertion make sure we are not logging complex types? 🤔

src/utils/dag/logStatus.js
pmuens: Nice catch and good idea!

src/utils/dag/logStatus.test.js
pmuens: Yes, that was also something I thought about after pushing. I'll update that :+1

src/utils/dag/logStatus.js
pmuens: Unfortunately `isType` doesn't work since it checks for a Type but not for an instance of a Type which is what we're dealing with here.

Additionally smth. like `isComponent` doesn't work since the `Provier` which is passed in via inputs most of the time is not a `Component`, but an `Object`.

src/utils/dag/logStatus.test.js
pmuens: Just updated the test so that we have a clear assertion and check to see that we don't output a message about Parameters when we only use the `Provider` as an input.

pmuens:
@eahefnawy thanks again for the review. I just updated the PR according to your review.

Additionally I added comments to your review comments for clarification.

PR #379, opened 2018-12-10
Add sync method to AwsDynamoDb component

pmuens:
Adds the `sync` method @eahefnawy implemented recently to the `AwsDynamoDb` component so that we check whether the table has already been removed before proceeding with the deployment / removal.

**Note:** This PR needs https://github.com/serverless/components/pull/380 to work correctly.


pmuens:
just updated the functionality so that all the remote state data is retrieved via `describeTable` and stored in the state file when running `sync`.

Furthermore I added unit tests which cover this functioanlity.

registry/AwsDynamoDb/src/utils/createTable.js
```diff
@@ -1,4 +1,4 @@
-const createTable = async (properties) => {
+async function createTable(properties) {
```
eahefnawy: any reason why you're switching off of arrow functions?

pmuens: Yes, I've decieded to stick with the `function` usage for the last couple of months.

The reason being that IMHO it's way more readable to have something like `async function foo() {}` compared to `const foo = async () => {}`. An additional benefit is that your code editor / IDE will highlight the code in a differnent way, adding yet another benefit of using the `function` syntax.

PR #380, opened 2018-12-10
Auto resolve credentials in AwsProvider

pmuens:
This needs to be in place so that the component `sync` function works since we're using `provider` there.

### DIFF
--- a/registry/AwsProvider/src/index.test.js
+++ b/registry/AwsProvider/src/index.test.js
@@ -2,6 +2,9 @@ import AWS from 'aws-sdk'
 import path from 'path'
 import { createTestContext } from '../../../test'
 
+// TODO: move this into @serverless/utils ?!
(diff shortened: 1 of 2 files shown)


registry/AwsProvider/src/index.test.js
```diff
@@ -2,6 +2,9 @@ import AWS from 'aws-sdk'
 import path from 'path'
 import { createTestContext } from '../../../test'
 
+// TODO: move this into @serverless/utils ?!
+import newVariable from '../../../src/utils/variable/newVariable'
```
eahefnawy: hmmm shouldn't this import from `dist`?

pmuens: Yes, usually it should if it's not a test (since Jest transpiles via Babel on the fly).

I'll switch it back to `dist` for consistency sake...

PR #382, opened 2018-12-12
Lambda layers

dschep:
## What has been implemented?

Added and `AwsLambdaLayerVersion` component.

Also refactored boiler plate in `loadTypeMetaFromName`

## Steps to verify

create a service using it. minimum config is `content` key pointing to a path

## Todos:

* [x] Write tests
* [x] Run Prettier


registry/AwsLambdaLayerVersion/src/index.js
```diff
@@ -0,0 +1,175 @@
...
+        ((this.retain === false && configChanged) ||
+          (prevInstance.layerName !== currentConfig.layerName && this.retain !== true))
+      ) {
+        return 'replace'
+      } else if (!prevInstance || configChanged) {
+        return 'deploy'
+      }
+    }
+
+    getId() {
```
pmuens: Personal preference, but why isn't this called `getArn`?

registry/AwsLambdaLayerVersion/src/index.js
```diff
@@ -0,0 +1,175 @@
...
+
+      if (isArray(this.content)) {
+        inputDirPath = this.content[0] // first item is path to content dir
+      }
+
+      const outputFileName = `${this.instanceId}-${Date.now()}.zip`
+      const outputFilePath = path.join(tmpdir(), outputFileName)
+
+      await packDir(inputDirPath, outputFilePath)
+      this.zip = readFileSync(outputFilePath)
```
pmuens: Same comment about `async` usage.

registry/AwsLambdaLayerVersion/src/index.js
```diff
@@ -0,0 +1,175 @@
...
+      const Lambda = new AWS.Lambda()
+
+      if (retain === true) {
+        context.log(`Skipping removing(per retention policy) Lambda Layer Version: ${layerName}`)
+        return
+      }
+
+      context.log(`Removing Lambda Layer Versions: ${layerName}`)
+
+      for (const version of versions) {
```
pmuens: Using `async` / `await` inside of `for` / `forEach` loops can be tricky (see https://serverless.com/blog/common-node8-mistakes-in-lambda/ and https://codeburst.io/javascript-async-await-with-foreach-b6ba62bbf404).

We might want to use `Promise.all` with a `catch` here (or Bluebirds `Promise.mapSeries` if sequential processing is needed).

src/utils/type/loadTypeMetaFromName.js
```diff
@@ -3,50 +3,46 @@ import { resolve } from 'path'
...
-    '..',
-    'registry',
-    'AwsCloudWatchEventsRule'
-  ),
-  TwilioPhoneNumber: resolve(__dirname, '..', '..', '..', 'registry', 'TwilioPhoneNumber'),
-  AwsSnsTopic: resolve(__dirname, '..', '..', '..', 'registry', 'AwsSnsTopic'),
-  AwsSnsSubscription: resolve(__dirname, '..', '..', '..', 'registry', 'AwsSnsSubscription'),
-  AwsDynamoDb: resolve(__dirname, '..', '..', '..', 'registry', 'AwsDynamoDb')
-}
+export const NATIVE_NAMES = componentsFromRegistryByName([
```
pmuens: 👏

registry/AwsLambdaLayerVersion/src/index.js
```diff
@@ -0,0 +1,175 @@
...
+      super.hydrate(prevInstance)
+      this.arn = get('arn', prevInstance)
+      this.zip = get('zip', prevInstance)
+      this.layerVersion = get('layerVersion', prevInstance)
+      this.versions = get('versions', prevInstance)
+    }
+
+    async shouldDeploy(prevInstance, context) {
+      if (isArchivePath(this.content)) {
+        this.zip = readFileSync(this.content)
```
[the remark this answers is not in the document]
eahefnawy: the latest `fs` should support promises by default now. Also I believe the utils repo has an async version

registry/AwsLambdaLayerVersion/src/index.js
eahefnawy: @pmuens I believe @dschep was inspired by `AwsLambdaFunction`, in which case it's called `getId` to keep it provider agnostic when it's called by the `Function` type. However in then case of Layers, it's an AWS only feature, so I agree with you.

registry/AwsLambdaLayerVersion/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "@serverless-components/aws-lambda-layer-version",
+  "version": "0.1.0",
```
eahefnawy: this should be 0.3.0 as well

registry/AwsLambdaLayerVersion/src/index.js
```diff
@@ -0,0 +1,175 @@
...
+      provider = resolve(provider)
+      const AWS = provider.getSdk()
+      const Lambda = new AWS.Lambda()
+
+      const { LayerVersions } = await Lambda.listLayerVersions({
+        LayerName: resolve(this.layerName)
+      }).promise()
+      if (isEmpty(LayerVersions)) {
+        return 'removed'
+      }
```
eahefnawy: that's interesting. So if a layer does not exist, it does not throw some sort of "resource does not exist" error? 🤔

registry/AwsLambdaLayerVersion/src/index.js
```diff
@@ -0,0 +1,175 @@
...
+        return
+      }
+
+      context.log(`Removing Lambda Layer Versions: ${layerName}`)
+
+      for (const version of versions) {
+        try {
+          await deleteLayerVersion(Lambda, layerName, version)
+        } catch (error) {
+          if (error.content !== 'ResourceNotFoundException') {
```
eahefnawy: continuing on my point above....deleting a layer throws a resource not found exception?

src/utils/type/loadTypeMetaFromName.js
```diff
@@ -3,50 +3,46 @@ import { resolve } from 'path'
...
 
+const componentsFromRegistryByName = (names) =>
+  names.reduce(
+    (obj, name) => ({
+      ...obj,
+      [name]: resolve(__dirname, '..', '..', '..', 'registry', name)
+    }),
+    {}
+  )
+
```
eahefnawy: awesome!! Thanks for that! 😄

registry/AwsLambdaLayerVersion/src/index.js
```diff
@@ -0,0 +1,175 @@
...
+  const params = { LayerName: layerName, VersionNumber: version }
+  await Lambda.deleteLayerVersion(params).promise()
+}
+
+const AwsLambdaLayerVersion = async (SuperClass) => {
+  return class extends SuperClass {
+    hydrate(prevInstance) {
+      super.hydrate(prevInstance)
+      this.arn = get('arn', prevInstance)
+      this.zip = get('zip', prevInstance)
```
eahefnawy: I don't think hydrating the zip binary is necessary. I don't believe we serialize binaries anyway. cc @brianneisler

registry/AwsLambdaLayerVersion/src/index.js
dschep: Correct. I also thought this was odd.

registry/AwsLambdaLayerVersion/src/index.js
dschep: yup.. doesn't it also just succeeds

registry/AwsLambdaLayerVersion/src/index.js
dschep: oops. forgot that. When's js getting `async for`?

registry/AwsLambdaLayerVersion/src/index.js
dschep: Correct. I simply copied that from lambda func

__mocks__/aws-sdk.js
```diff
@@ -109,6 +109,23 @@ const mocks = {
     }
     return Promise.resolve()
   }),
+  publishLayerVersionMock: jest.fn().mockReturnValue({ LayerArn: 'abc:zxc' }),
+  deleteLayerVersionMock: jest.fn().mockImplementation((params) => {
+    if (params.FunctionName === 'already-removed-layer') {
+      const error = new Error()
+      error.code = 'ResourceNotFoundException'
```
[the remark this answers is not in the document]
dschep: Just tested how the func actually behaves. it just returns an empty object with no error when the layer version doesn't exist

__mocks__/aws-sdk.js
```diff
@@ -109,6 +109,23 @@ const mocks = {
...
+      return Promise.reject(error)
+    }
+    return Promise.resolve()
+  }),
+  listLayerVersionsMock: jest.fn().mockImplementation((params) => {
+    if (params.FunctionName === 'already-removed-layer') {
```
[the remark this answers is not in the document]
dschep: also returns empty response rather than throwing an error actaully. updating to reflect that.

registry/AwsLambdaLayerVersion/serverless.yml
```diff
@@ -0,0 +1,64 @@
...
+  layerDescription:
+    type: string
+    required: false
+    displayName: Lambda layer description
+    description: |
+      A short, user-defined layer description.
+      Lambda does not use this value. Assign a meaningful description as you see fit
+    example: The layer that does XYZ to ABC
+  content:
+    type: string || array || buffer
```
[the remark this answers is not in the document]
pmuens: According to the Raml spec this should be `string | array | buffer` (see [union types](https://github.com/raml-org/raml-spec/blob/master/versions/raml-10/raml-10.md#union-type))

registry/AwsLambdaLayerVersion/serverless.yml
```diff
@@ -0,0 +1,64 @@
...
+    description: Runtimei the layer is compatible with.
+    example: [nodejs8.10]
+  licenseInfo:
+    type: string
+    required: false
+    displayName: License Info
+    description: License Information for the layer
+    example: Apache-2.0
+  retain:
+    type: boolean || string
```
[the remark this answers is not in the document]
pmuens: According to the Raml spec this should be `string | array | buffer` (see [union types](https://github.com/raml-org/raml-spec/blob/master/versions/raml-10/raml-10.md#union-type))

registry/AwsLambdaLayerVersion/src/index.js
```diff
@@ -0,0 +1,169 @@
...
+    LicenseInfo: licenseInfo,
+    CompatibleRuntimes: compatibleRuntimes
+  }
+
+  return await Lambda.publishLayerVersion(params).promise()
+}
+
+const deleteLayerVersion = async (Lambda, layerName, version) => {
+  const params = { LayerName: layerName, VersionNumber: version }
+  await Lambda.deleteLayerVersion(params).promise()
```
brianneisler: Maybe just return here instead of await?

registry/AwsLambdaLayerVersion/src/index.js
```diff
@@ -0,0 +1,169 @@
...
+      this.arn = get('arn', prevInstance)
+      this.layerVersion = get('layerVersion', prevInstance)
+      this.versions = get('versions', prevInstance)
+    }
+
+    async shouldDeploy(prevInstance, context) {
+      if (isArchivePath(this.content)) {
+        this.zip = await readFile(this.content)
+      } else {
+        await this.pack(context)
```
brianneisler: This is fine for now, but it might be worth adding some caching here eventually to prevent from having to pack on every deployment.

registry/AwsLambdaLayerVersion/src/index.js
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    async pack() {
+      let inputDirPath = this.content
+
+      if (isArray(this.content)) {
+        inputDirPath = this.content[0] // first item is path to content dir
+      }
+
+      const outputFileName = `${this.instanceId}-${Date.now()}.zip`
```
brianneisler: I believe adding the timestamp here will break the caching check and result in a new hash on every deployment. It would be better to use a hash derived from the folder contents and use that as part of the name. This will ensure uniqueness on a per deployment basis and that there is only overlap when the code is the same.

registry/AwsLambdaLayerVersion/src/index.js
```diff
@@ -0,0 +1,169 @@
...
+
+      const { LayerVersions } = await Lambda.listLayerVersions({
+        LayerName: resolve(this.layerName)
+      }).promise()
+      if (isEmpty(LayerVersions)) {
+        return 'removed'
+      }
+      this.versions = LayerVersions.map(({ Version }) => Version)
+      this.layerVersion = Math.max(...this.versions)
+      this.arn = LayerVersions.filter(
```
brianneisler: `find` method from utils does exactly this but is more performant as it stops the iteration on the first match.

registry/AwsLambdaLayerVersion/src/index.js
```diff
@@ -0,0 +1,169 @@
...
+      const Lambda = new AWS.Lambda()
+
+      if (retain === true) {
+        context.log(`Skipping removing(per retention policy) Lambda Layer Version: ${layerName}`)
+        return
+      }
+
+      context.log(`Removing Lambda Layer Versions: ${layerName}`)
+
+      await Promise.all(versions.map((version) => deleteLayerVersion(Lambda, layerName, version)))
```
brianneisler: Not necessary, but a note that the `mapAll` method from utils will auto upgrade to async so you don't need to use `Promise.all`

```
await mapAll((version) => deleteLayerVersion(Lambda, layerName, version), versions)
```

src/utils/type/loadTypeMetaFromName.js
```diff
@@ -3,50 +3,46 @@ import { resolve } from 'path'
 import errorUnknownTypeName from './errorUnknownTypeName'
 import loadTypeMetaFromPath from './loadTypeMetaFromPath'
 
+const componentsFromRegistryByName = (names) =>
```
brianneisler: so much cleaner, thanks!

registry/AwsLambdaLayerVersion/src/index.js
brianneisler: As soon as async iterators land https://www.codementor.io/tiagolopesferreira/asynchronous-iterators-in-javascript-jl1yg8la1

registry/AwsLambdaLayerVersion/src/index.js
dschep: Nope, filename is never part of the object (only the contents and the checksum of its contents). I've manually tested this too, it doesn't push a new layer unless the content or config changes.

PR #383, opened 2018-12-13
Refactor AwsApiGateway / Fix RestApi integration

eahefnawy:
It also fixes a bug I discovered when updating the API where changes are merged rather than overwritten.


registry/AwsApiGateway/serverless.yml
```diff
@@ -10,30 +10,18 @@ license: Apache-2.0
...
+  swaggerTemplate:
     type: object
-    required: true
-    description: Declaration of routes
-  role:
-    type: AwsIamRole
-    required: true
-    description: AWS Arn role for API gateway to assume
-  apiName:
+    required: false # setting this to false otherwise the core will print the entire template on every deployment
```
pmuens: Let's rather switch that to true and tackle the logging issue ASAP. Otherwise the integrity of this component is harmed.

registry/AwsApiGateway/src/index.js
```diff
@@ -1,197 +1,126 @@
...
 const AwsApiGateway = function(SuperClass) {
   return class extends SuperClass {
     async construct(inputs, context) {
       await super.construct(inputs, context)
-
-      this.provider = resolvable(() => or(inputs.provider, context.get('provider')))
-      this.apiName = resolvable(() => or(inputs.apiName, `apig-${this.instanceId}`))
-      this.role = inputs.role
-      this.routes = inputs.routes
+      this.provider = inputs.provider
```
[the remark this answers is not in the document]
eahefnawy: yeah but we need to `resolve` method in the constructor for the stage

registry/AwsApiGateway/serverless.yml
eahefnawy: 👍

registry/AwsApiGateway/src/index.js
```diff
@@ -1,197 +1,126 @@
...
       }
+
+      try {
+        await APIGateway.getExport(params).promise()
+
+        // Eslam:
+        // practically in the real world, the exported swagger template object
+        // will always be different than the imported one in small unpredictable details
+        // imo in this particular case, the sync should only check the existence of the API
+        // but not update the swagger template, and instead rely on the local template for idempotency
```
pmuens: Yes, I agree with that. In that case we could use `getRestApi` to check for the presence of the deployed Rest API. Not entirely sure what `getExport` returns at the end of the day, but this could be larger in terms of data sent over the wire compared to a simple `getRestApi` query.

registry/RestApi/src/index.js
```diff
@@ -1,130 +1,43 @@
...
-      childComponents.push(this.gateway)
+      // eslint does not allow destructuring because this is a reserved word...
+      const roleArn = newVariable('${this.children.role.arn}', { this: this })
+      const flatRoutes = flattenRoutes(resolve(this.routes))
+      const normalizedRoutes = normalizeRoutes(flatRoutes)
+      const swaggerTemplate = getSwaggerDefinition(resolve(this.apiName), roleArn, normalizedRoutes)
 
-      return childComponents
+      return {
+        role: await context.construct(iamComponent, {
```
pmuens: Can we also get rid of this `await` since we're not using it for the gateway below?

pmuens:
Just finalized the PR by adding tests for the `RestApi` and `AwsApiGateway` component.

This PR can be considered done and should be in a mergeable state.

/cc @eahefnawy @brianneisler

PR #390, opened 2019-01-04
Fix examples directory / refactor SimpleService example

pmuens:
Fixes the examples directory to only include a `SimpleService` and RFCs.

We shouldn't include tons of examples here since they'll only get stale and will confuse people who want to try them out.

I tried to add the `RestApi` component to the `SimpleService`, but it failed invoking the function, so I removed it. We can re-add it once we're done with all the API Gateway / REST API related changes.

/cc @brianneisler @eahefnawy


examples/SimpleService/serverless.yml
```diff
@@ -0,0 +1,38 @@
...
+extends: Service
+
+credentials:
+  type: EnvironmentBasedConfiguration
+  inputs:
+    variables:
+      - AWS_ACCESS_KEY_ID
+      - AWS_SECRET_ACCESS_KEY
+      - AWS_SECURITY_TOKEN
+      - AWS_SESSION_TOKEN
```
eahefnawy: Is this is how we're encouraging users to set up credentials?

I'm honestly not a big fan of this component, I think it really complicates that "simple" service, specially that we can do it without it. With components like these, our serverless.yml is starting to look like CF.

examples/SimpleService/serverless.yml
```diff
@@ -0,0 +1,38 @@
...
+
+providers:
+  aws:
+    type: AwsProvider
+    inputs:
+      credentials:
+        accessKeyId: ${this.credentials.values.AWS_ACCESS_KEY_ID}
+        secretAccessKey: ${this.credentials.values.AWS_SECRET_ACCESS_KEY}
+        securityToken: ${this.credentials.values.AWS_SECURITY_TOKEN}
+        sessionToken: ${this.credentials.values.AWS_SESSION_TOKEN}
```
eahefnawy: I dislike it even more when I look at all those long strings :/ ... we should really minimize the configuration, specially for the simple service example. I think the fact that the AwsProvider component automatically picks it up is a great start.

pmuens:
Hey @eahefnawy thanks for the review! 👍

Yes, I agree that the `env` component is a hack and very verbose. However using this seems to be the only way for me to work with examples since I'm using our companys MFA protected account. I'm pretty sure that lots of enterprise users who dig into components also have guarded AWS accounts (@shortjared knows more here).

I'd love to keep the examples we ship focues and simple. My hunch is that as soon as we add more and more examples we forget to update them, causing confusion (a situation we're already facing right now). The best of both worlds would be if we could utilize our examples as integration tests (however that might look like).

The TL;DR is that I have no idea if we should ditch or keep the env component for now. Ditching it might cause enterprise users with AWS MFA accounts to face issues exploring components.

/cc @brianneisler

eahefnawy:
hmmm @pmuens good point. I didn't think about MFA as I'm using my personal account. But it's a valid concern. 👍 

This PR is G2M then

pmuens:
> hmmm @pmuens good point. I didn't think about MFA as I'm using my personal account. But it's a valid concern. 👍 ... if removing it blocks you, then let's keep it.
> This PR is G2M then

Awesome! Thanks for the thorough PR review @eahefnawy 👍

Can you approve the PR so that I can merge it? Thanks

PR #392, opened 2019-01-09
Remove unnecessary logStatus reportings

pmuens:
This PR fixes the verbose status reportings `logStatus` performs when the graph is walked.

The implementation defines an array of `rootComponents` which each node / component is checked against. Status reports are only logged for nodes / components that don't extend one of the `rootComponets` (e.g. the `Service` the user has defined when working on a project).

This is one of many ways to solve this...

/cc @brianneisler @eahefnawy


pmuens:
Closing this PR because of an upcoming update...

PR #416, opened 2019-03-27
Add support for @latest as a component version number

pmuens:
This adds support for component declarations as `AwsLambda@latest`, therefore removing the need to specify a version number.

Using `LATEST` will always pull from GitHubs `master`.


pmuens:
Closing since we're not using GitHub anymore...

PR #417, opened 2019-03-28
add custom chat app example

ac360:
## What has been implemented?

Closes #XXXXX

## Steps to verify

* Run X
* ...

## Todos:

* [ ] Write tests
* [ ] Run Prettier


templates/chat-app-custom/backend/package.json
```diff
@@ -0,0 +1,14 @@
...
+  "version": "1.0.0",
+  "description": "",
+  "main": "socket.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "aws-sdk": "^2.429.0"
```
eahefnawy: this shouldn't be required, and probably slows down deployment a lot! Any particular reason why you added it?

eahefnawy:
Also, imo this should be THE chatApp, not another "custom" chatApp. The one we had before doesn't really do much :)

PR #442, opened 2019-04-27
Fix local component path reduce

laardee:
## What has been implemented?

Fixes local component path reduce and some read me fixes


ac360:
Thank you!  I run into this issue all of the time.

laardee:
@ac360 np, this started to bug me also a little bit because I'm using the SNS topic and SNS subscription components (which I'm implementing) from local directories.

PR #443, opened 2019-04-29
Auto component removal

pmuens:
Brings back the auto-removal of components (when they're removed from `serverless.yml`).


eahefnawy:
Thanks @pmuens

PR #447, opened 2019-05-12
cli options improvements

agutoli:
## What has been implemented?
conponents cli improvements

## Examples 

Screen Shot 2019-05-12 at 7 01 22 pm
Screen Shot 2019-05-12 at 7 01 41 pm
Screen Shot 2019-05-12 at 7 01 57 pm
Screen Shot 2019-05-12 at 7 02 16 pm
Screen Shot 2019-05-12 at 7 03 09 pm

### DIFF
--- a/bin/cli
+++ b/bin/cli
@@ -17,7 +17,6 @@ const parser = new ArgumentParser({
 parser.addArgument('method', {
   action: 'store',
   nargs: "?",
-  choices: ['deploy', 'remove'],
   defaultValue: null,
   help: 'Methods'
 })
--- a/bin/cli
+++ b/bin/cli
@@ -3,19 +3,51 @@
 /*
  * Serverless Components – Command Line Interface
  */
+const path = require('path');
+const { ArgumentParser } = require('argparse');
+const { version } = require(path.resolve(path.join(__dirname, '../package.json')));
+
+const parser = new ArgumentParser({
+  version: version,
+  addHelp:true,
+  prog: "components",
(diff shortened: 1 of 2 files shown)


bin/cli
```diff
@@ -3,20 +3,52 @@
 /*
  * Serverless Components – Command Line Interface
  */
+const path = require('path');
+const { ArgumentParser } = require('argparse');
+const { version } = require(path.resolve(path.join(__dirname, '../package.json')));
```
eahefnawy: thanks for that!

bin/cli
```diff
@@ -3,20 +3,52 @@
...
+  version: version,
+  addHelp:true,
+  prog: "components",
+  description: 'Serverless components usage'
+});
+
+parser.addArgument('method', {
+  action: 'store',
+  nargs: "?",
+  choices: ['deploy', 'remove'],
```
eahefnawy: hmmm what do these choices refer to exactly? The components core is pretty agnostic when it comes to components methods. Each component only needs the `default` method as a requirement. All else is optional. So not sure how `deploy` fits in here?

Am I missing something? 🤔

agutoli: This is just an example, we can just remove it and keep it agnostic. The central idea is refactoring it to use a cli library.

agutoli: @eahefnawy I just remove the options.

eahefnawy:
LG2M

Thanks @agutoli :raised_hands:

PR #464, opened 2019-08-02
Document that components need to be exported

mnapoli:
## What has been implemented?

Fix the "Getting started" examples. In my experience I had to export the component to make things work.

I hope this is a change that makes sense, please review as I am no NodeJS expert :)

## Todos:

* [ ] Write tests
* [ ] Run Prettier


eahefnawy:
Yep! that's why! Thanks a lot for the PR @mnapoli

PR #509, opened 2019-10-13
Consume all method arguments when calling built-in methods.

Pavel910:
This PR fixes a bug with built-in class methods arguments forwarding. The way they were implemented would only consume the first argument which would cause wrong behavior on methods like `load` (which accepts 2 arguments).

This fix is required for PR [serverless/template#8](https://github.com/serverless/template/pull/8) to work properly (without it, custom methods will not have access to the component state).

@eahefnawy as always, please review and send feedback if any changes are necessary :)


eahefnawy:
Thanks @Pavel910 ... LG2M! 🎉

PR #555, opened 2020-02-20
fix: global proxy not work bug

yugasun:
Sometimes, users can not access to `https://registry.npmjs.org` directly, so they add global proxies for it. But  `got` lib (`package-json` depends on it) don't respect it, so they will get a `ECONNREFUSED` error. Refer to this issue https://github.com/sindresorhus/got/issues/560, so we need [global-tunnel-ng](https://github.com/np-maintain/global-tunnel).

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
   "license": "MIT",
   "dependencies": {
     "fs-extra": "^7.0.1",
+    "global-tunnel-ng": "^2.7.1",
     "js-yaml": "^3.13.1",
     "package-json": "^6.3.0",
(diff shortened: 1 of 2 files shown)


pmuens:
Hey @yugasun thanks for working on this.

I'm unfamiliar with those issues but it seems like the solution fixes the problem. Do you think that https://github.com/sindresorhus/got/issues/560#issuecomment-435704556 can happen for users working with Components?

yugasun:
@pmuens sure, I can produce it on my internal develop machine.

PR #570, opened 2020-04-02
fix: respect global proxy

yugasun:
Sometimes, users can not access to `https://registry.npmjs.org` directly, so they add global proxies for it. But  `got` lib (`package-json` depends on it) don't respect it, so they will get a `ECONNREFUSED` error. Refer to this issue https://github.com/sindresorhus/got/issues/560, so we need [global-tunnel-ng](https://github.com/np-maintain/global-tunnel).

### DIFF
--- a/src/utils/download.js
+++ b/src/utils/download.js
@@ -10,16 +10,21 @@ const semver = require('semver')
 const util = require('util')
 const exec = util.promisify(require('child_process').exec)
 const dirExists = require('./fs/dirExists')
-const globalTunnel = require('global-tunnel-ng')
+
+const MAJOR_NODEJS_VERSION = parseInt(process.version.slice(1).split('.')[0], 10)
+if (MAJOR_NODEJS_VERSION >= 10) {
+  // `global-agent` works with Node.js v10 and above.
+  require('global-agent').bootstrap()
+  http.globalAgent.keepAlive = true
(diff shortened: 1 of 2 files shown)


src/utils/download.js
```diff
@@ -1,14 +1,25 @@
 const os = require('os')
 const path = require('path')
+const http = require('http')
+const https = require('https')
 const { ensureDir, remove } = require('fs-extra')
 const packageJson = require('package-json')
+const registryUrl = require('registry-url')
```
medikoo, line 7: We need to list it in `package.json`

src/utils/download.js
```diff
@@ -1,14 +1,25 @@
...
 const semver = require('semver')
 // const BbPromise = require('bluebird')
 const util = require('util')
 const exec = util.promisify(require('child_process').exec)
 const dirExists = require('./fs/dirExists')
+const globalTunnel = require('global-tunnel-ng')
```
medikoo, line 13: In package documentation it says it won't work on Node.js v11 and above, while most of us uses v12 or v13 already.

Can we use an alternative that works on all modern Node.js versions?

yugasun:
@medikoo All changes done.

medikoo:
@eahefnawy can you give it a second look?

PR #578, opened 2020-04-07
Fix process.env resolution

medikoo:
When working on Tencent commands, we've approached some issues, and after investigation it appeared that it's caused by fact of environment variables not being resolved from `.env` files

Ideally `.env` files should be resolved at earliest stage of a program, and further logic should simply refer to `process.env` for environment variables values.

This PR ensures it's the case.

As there's no test suite, I've tested manually by running few component commands. All worked without issues.

### DIFF
--- a/src/cli/commands/utils.js
+++ b/src/cli/commands/utils.js
@@ -12,7 +12,7 @@ const {
   refreshToken,
   listTenants
 } = require('@serverless/platform-sdk')
-const { fileExistsSync, readFileSync, resolveInputVariables } = require('../utils')
+const { fileExistsSync, loadInstanceConfig, resolveInputVariables } = require('../utils')
 
 const getDefaultOrgName = async () => {
   const res = readConfigFile()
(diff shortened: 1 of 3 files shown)


src/cli/commands/utils.js
```diff
@@ -115,48 +99,13 @@ const loadInstanceCredentials = (stage) => {
...
-    isYaml = true
-  }
-  if (fileExistsSync(yamlFilePath)) {
-    filePath = yamlFilePath
-    isYaml = true
-  }
-  if (fileExistsSync(jsonFilePath)) {
-    filePath = jsonFilePath
-  }
+const loadAwsInstanceConfig = async (directoryPath) => {
```
ac360, line 102: Components is multi-vendor and we don't want to create vendor-specific methods like this.

medikoo: This is in AWS specific directory (note we separated commands into vendor specific folders)

ac360: There is no AWS-specific directory is my point.  There can be Tencent-specific stuff, but everything else must be vendor agnostic.

medikoo: Altough indeed I see there's also resolution for Google env vars and Tencent. Can it work on Google?

And I take for Tencent it's dead logic

medikoo: Ok so I'll rename it to `loadVendorInstance Config`

medikoo: Updated.

Note this is internal name, not exposed outside of module. It's just to not colide with more generic `loadInstanceConfig` which doesn't attempt to resolve vendor specific info.

PR #581, opened 2020-04-07
Change README environment var references

futuraprime:
## What has been implemented?

Potentially addresses #580. (I am not sure if `env.` is the desired behavior, but this change seems small enough it made sense to just submit a fix for the documentation directly; if you prefer to change the behavior, ignore this.)

## Steps to verify

* This is just a documentation change. You can attempt to deploy an app using `{env:FOO}` in the configuration file to verify that the syntax does not work, while `{env.FOO}` does work (at least with serverless 1.67.1).

## Todos:

* [ X ] Write / update documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -410,7 +410,7 @@ stage: prod
 
 inputs:
   name: ${org}-${stage}-${app}-${name} # Results in "acme-prod-ecommerce-rest-api"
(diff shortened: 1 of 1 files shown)


eahefnawy:
@futuraprime I believe you're still using v1 of Components. The master is now at v2. Here are the docs you're looking for in the `v1` branch:
https://github.com/serverless/components/tree/v1#environment-variables

PR #585, opened 2020-04-09
update name of platform deps in China

hkbarton:
replace all deps regarding tencent, remove word `tencent`

### DIFF
--- a/src/cli/commands-cn/dev.js
+++ b/src/cli/commands-cn/dev.js
@@ -6,7 +6,7 @@ const path = require('path')
 const { Writable } = require('stream')
 const ansiEscapes = require('ansi-escapes')
 const chokidar = require('chokidar')
-const { ServerlessSDK, utils: tencentUtils } = require('@serverless/tencent-platform-client')
+const { ServerlessSDK, utils: chinaUtils } = require('@serverless/platform-client-china')
 const utils = require('./utils')
 
 class LogForwardingOutput extends Writable {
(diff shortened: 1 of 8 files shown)


src/cli/commands-cn/dev.js
```diff
@@ -6,7 +6,7 @@ const path = require('path')
 const { Writable } = require('stream')
 const ansiEscapes = require('ansi-escapes')
 const chokidar = require('chokidar')
-const { ServerlessSDK, utils: tencentUtils } = require('@serverless/tencent-platform-client')
+const { ServerlessSDK, utils: tencentUtils } = require('@serverless/platform-client-china')
```
eahefnawy, line 9: not sure of the context of this whole change, but don't we also wanna rename `tencentUtils`?

hkbarton: oh, good catch, let me change this name too

src/cli/commands-cn/dev.js
```diff
@@ -84,9 +84,9 @@ async function updateDeploymentStatus(cli, instanceInfo, startDebug) {
         state: { lambdaArn, region }
       } = instanceInfo
       if (lambdaArn && region) {
-        await tencentUtils.stopTencentRemoteLogAndDebug(lambdaArn, region, cliEventCallback)
+        await chinaUtils.stopTencentRemoteLogAndDebug(lambdaArn, region, cliEventCallback)
         if (startDebug) {
-          await tencentUtils.startTencentRemoteLogAndDebug(lambdaArn, region, cliEventCallback)
+          await chinaUtils.startTencentRemoteLogAndDebug(lambdaArn, region, cliEventCallback)
```
eahefnawy, line 89: do we wanna rename these two? We'll have to dig into the actual package code though. Not sure if it's worth it. Just something to note

hkbarton: I think for startTencentRemoteLogAndDebug we should not update, since it’s real tencent specific logic, you can image maybe in the future we’ll have chinaUtils.startAliCloudRemoteLogAndDebug

PR #587, opened 2020-04-13
fix bug in CN commands of not resolving dist folder

hkbarton:
fix a bug in China commands that not resolving the dist folder, this causes an issue when running deployment outside of the project folder by specifying the deployment target folder.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "@serverless/inquirer": "^1.1.0",
     "@serverless/platform-client": "^0.25.4",
     "@serverless/platform-sdk": "^2.3.0",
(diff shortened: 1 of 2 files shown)


eahefnawy:
Is that an issue only in China or the AWS version as well? Because I'm guessing this portion of the code was copied over from the AWS version?

hkbarton:
@eahefnawy this is only related to China command. There is an experimental feature in CN command that Tencent wants to play out, you can specify the target folder of which includes `serverless.yml` like `sls deploy --target=/folder-include-serverless.yml`. This solves the issue that not resolving `dist` in inputs.

PR #589, opened 2020-04-14
Knative integration

pmuens:
## What has been implemented?

Adds Environment variables to integrate the new Components implementation with a Kubernetes cluster (which in turn runs Knative).

## Steps to verify

Trust me, you don't want to test this end-to-end...

## Todos:

* [x] Run Prettier

### DIFF
--- a/src/cli/commands/utils.js
+++ b/src/cli/commands/utils.js
@@ -48,6 +48,7 @@ const getDefaultOrgName = async () => {
  * Load credentials from a ".env" or ".env.[stage]" file
  * @param {*} stage
  */
+
 const loadInstanceCredentials = () => {
   // Known Provider Environment Variables and their SDK configuration properties
   const providers = {}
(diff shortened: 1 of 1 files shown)


pmuens:
@eahefnawy here's the PR which adds the necessary environment variables for the Kubernetes / Knative integration.

eahefnawy:
Let me know if this is super urgent, then I can cut a release, otherwise, we'll ship this along with some other things.

pmuens:
Thanks for merging! :+1:

> Let me know if this is super urgent, then I can cut a release, otherwise, we'll ship this along with some other things.

It seems like it's not super urgent right now. I'll let you know if this changes and we should cut a release...

fernando-mc:
@eahefnawy @medikoo would it be possible to actually cut a release for this? I'm working on a demo today that needs it.

PR #590, opened 2020-04-15
ensure serverless.yml inputs are for "deploy" only

ac360:
## What has been implemented?

`serverless.yml` `inputs` property was only meant for the `deploy` action, not `remove` or any custom actions.  `inputs` actions other than deploy need to be handled separately, via a method we have not yet considered.

This fixes the first problem and prevents `inputs` from being automatically submit for custom actions and especially `remove` actions, which users currently use more within the CLI.

cc @hkbarton

### DIFF
--- a/src/cli/commands/run.js
+++ b/src/cli/commands/run.js
@@ -105,12 +105,18 @@ module.exports = async (config, cli, command) => {
     // cli.log()
     // cli.log(`${chalk.grey(`Full details: ${dashboardUrl}`)}`)
   } else if (command === 'remove') {
-    // run remove
     cli.status('Removing', null, 'white')
+
+    // The "inputs" in serverless.yml are only for deploy.  Remove them for all other commands
+    instanceYaml.inputs = {}
+
     await sdk.remove(instanceYaml, instanceCredentials, options)
   } else {
     // run a custom method synchronously to receive outputs directly
     options.sync = true
+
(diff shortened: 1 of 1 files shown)


src/cli/commands/run.js
```diff
@@ -105,13 +105,19 @@ module.exports = async (config, cli, command) => {
...
+    // The "inputs" in serverless.yml are only for deploy.  Remove them for all other commands
+    instanceYaml.inputs = {}
+
     await sdk.remove(instanceYaml, instanceCredentials, options)
   } else {
     // run a custom method synchronously to receive outputs directly
     options.sync = true
 
+    // The "inputs" in serverless.yml are only for deploy.  Remove them for all other commands
+    instanceYaml.inputs = {}
```
pmuens, line 119: This would force a custom method to fetch inputs via state, correct?

How would I pass-in inputs which aren't stored in the state?

pmuens: Just figured that I'm using the Components I wrote as pure npm dependencies so that doesn't affect me.

Feel free to ignore the comment above.

ac360: Hi @pmuens, state is shared across all methods, so custom methods can always use that—as long as the component has been deployed and there is state available.  But it defeats the purpose of having separate methods that do separate things, if each can't have separate arguments (i.e. inputs).

How to pass in inputs for custom methods is something we have to figure out still via the CLI.  Via the SDK, it's available .

PR #593, opened 2020-04-17
Deploy multiple components with --all flag

eahefnawy:


hkbarton:
hi @eahefnawy could you add those in the parent utils which can be shared cross `commands` and `commands-cn`. I think those logic can be shared across both

eahefnawy:
@hkbarton will do 🙌

src/cli/commands-cn/runAll.js
```diff
@@ -0,0 +1,111 @@
...
+    }
+  })
+
+  // Prepare Options
+  const options = {}
+  options.debug = config.debug
+  options.dev = config.dev
+
+  // Connect to Serverless Platform Events, if in debug mode
+  if (options.debug) {
```
hkbarton, line 48: we don't have debug option support for Tencent yet, so maybe we can comment out this whole section

src/cli/commands-cn/utils.js
```diff
@@ -159,9 +159,70 @@ const loadInstanceCredentials = (stage) => {
...
+      if (template.stage !== null && template.stage !== instanceYaml.stage) {
+        throw new Error('Attempting to deploy multiple instances to multiple stages')
+      }
+
+      template.org = instanceYaml.org // eslint-disable-line
+      template.app = instanceYaml.app // eslint-disable-line
+      template.stage = instanceYaml.stage // eslint-disable-line
+
+      // update paths in inputs
+      if (instanceYaml.inputs.src) {
```
hkbarton, line 201: this section is not needed for commands-cn since all the path has been resolved in loadTencentInstanceConfig

src/cli/utils.js
```diff
@@ -424,6 +435,161 @@ const isProjectPath = async (inputPath) => {
...
+        inputs: template[key].inputs || {}
+      }
+    }
+  }
+
+  return allComponents
+}
+
+const setDependencies = (allComponents) => {
+  const regex = /\${output:(\w*[\w.-_]+)}/g
```
hkbarton, line 495: found a issue during the test regarding this regex, it can't match stuff like `${output:${stage}:${app}:fullstack-api-v2.apigw.url}`, so basically those characters don't work: `$ { } -`. Changed it to `/\${output:(\w*[-_\${}:\w.]+)}/g` works.

hkbarton:
@eahefnawy I fixed those issue being mentioned in the above comments and have a PR here https://github.com/serverless/components/pull/596 (this PR is for merging to this branch, not for master)

tinafangkunding:
Thanks for the hard-working @hkbarton and @eahefnawy, you are awesome!

eahefnawy:
@tinafangkunding Thanks Tina 😊

PR #599, opened 2020-04-21
Adds support for AWS credentials file

### LINKED ISSUE #597: Support AWS Credentials Conventions
The most common issue users face when using components is that they forget or they don't know how to set up the `.env` file. It's even more problematic as users create multiple instances because they have to create the `.env` file in each instance directory.

We should make the `.env` file optional if the user already has AWS credentials setup

eahefnawy:
Closes #597 

This enables users to deploy components without a `.env` file if they already have the AWS credentials in their how directory.

This implementation has no dependency on the aws-sdk, and respects the AWS rules & env vars.


src/cli/commands/utils.js
```diff
@@ -46,12 +52,53 @@ const getDefaultOrgName = async () => {
   return defaultOrgName
 }
 
+const loadAwsCredentials = () => {
```
ac360, line 55: @eahefnawy Components **must** have a great vendor-agnostic credentials handling experience, that treats all vendors as equals.  The CLI architecture should contain a pattern for vendor-agnostic credential handling that can easily scale to support new vendors, without requiring a re-architect in a few weeks when we add in another vendor.  It looks like like the opinion here is to use `loadInstanceCredentials` to fetch Credentials for all vendors, via separate functions that fetch credentials for each vendor.  This seems fine.  Though, could you write some comments in the codebase above these functions to make this pattern clear, so a person without experience or open-source contributor can submit a PR to add a new vendor, while following our scalable pattern for doing so?

eahefnawy: Ah good catch. Done.

eahefnawy:
> we cannot really ensure complete AWS authentication experience that way. What about automatic authentication via EC2 Metadata Service? What about support for Web Identity tokens? I believe to ensure solid authentication experience in all cases we need to fallback to aws-sdk, as it's the only way to guarantee all methods will work.

@medikoo I don't think we should worry about more advanced aws authentication methods. At least at this point, components are targeting people who are new to Serverless and even the cloud. If they are advanced cloud users, they probably already have their credentials set locally and also wouldn't have issues setting them up in `.env` files.

> I think the right way to go, would be to create @serverless/aws-component (or similar) as a core utils/logic base for any component that operates with AWS provider. That will be naturally used by users wishing to rely on this provider, and over there I believe it's totally fine to have aws-sdk dependency and ensure that AWS authentication works naturally.

Could you clarify that a bit? Credentials handling cannot be a component (or even a component dependency) because it needs to run locally via the CLI, and for best UX for each provider, we also have to support their built-in authentication methods, just like we're doing with Tencent.

medikoo:
> Could you clarify that a bit? Credentials handling cannot be a component (or even a component dependency) because it needs to run locally via the CLI

If I understand correctly, situation is that we want to allow users to deploy component instances to already deployed components as e.g. `aws-lambda`.

And in such situation user doesn't install `aws-lambda` locally, but purely via `serverless.yml` configuration signals the intention to deploy directly to it.
In result there's no AWS dedicated package/utility installed on user side (in scope of a project), which allows to provide AWS authentication logic to (ideally) vendor agnostic CLI.

In given situation, if I understand it correctly, if we don't want user to install platform specific dedicated separate utilities set (to be used by our CLI), we're not really capable to propose a _vendor agnostic_ CLI.

Best we can do is to provide a _multi vendor_ CLI (which imo differs from being _vendor agnostic_), that comes in with bundled logic, dedicated for each supported platform, (to handle authentication and whatever else we may be needed in generic layer).

Note that with such approach, support for each vendor needs to be bundled into CLI, there seems no way to provide a support for some _vendor_ X without extending the CLI.

Anyway if such approach is what we agreed on, then indeed what you propose looks as a good start (I got a bit tricked thinking we're after pure _vendor agnostic_ solution)

eahefnawy:
@medikoo thanks for clarifying. Yeah, you're totally right. That's the dilemma we are in. Ultimately what we settled on is that credentials handling will probably be the simplest provider-specific logic that we will need to maintain and the most critical. All the rest could go into provider-specific components. So yeah I guess you're right, we should call it multi-provider CLI 😅

PR #606, opened 2020-04-27
enable putting .env files in the parent directory

ac360:
## What has been implemented?

When using `deploy --all` you can deploy everything together.  However, immediately after that, you will want to deploy things separately, to provision updates without risking the application as a whole.  The problem you may run into then is that your `.env` file is only in the parent directory, and you will need to copy it to every subfolder. 

This change enables looking 1 folder up for a `.env` file when deploying a specific component, which is helpful in the use-case above.

## Steps to verify

* Put a `.env` file in a parent directory, above a component instance you wish to deploy.
* Deploy the component instance in the child directory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ Next, log in to enable deployment and saving state in the cloud:
 $ serverless login
 ```
 
-Lastly, enter your cloud provider credentials into a `.env` file within the folder that contains your `serverless.yml`.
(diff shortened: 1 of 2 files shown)


yugasun:
For me, it's a very useful feature.

hkbarton:
this looks good @ac360

tinafangkunding:
Yes, very helpful for multi-components management, and users can deploy a single component as they want.

PR #619, opened 2020-05-01
Run components login in template directory

eahefnawy:
Runs components if running in a directory with templates.

### DIFF
--- a/src/cli/utils.js
+++ b/src/cli/utils.js
@@ -439,11 +439,23 @@ const isProjectPath = async (inputPath) => {
   return false;
 };
 
-const runningTemplate = (config, command) => {
-  if (config.all && ['deploy', 'remove'].includes(command)) {
-    return true;
+const runningTemplate = (root) => {
+  const directories = fse
+    .readdirSync(root)
+    .filter((f) => fse.statSync(path.join(root, f)).isDirectory());
+
+  let componentDirectoryFound = false;
+  for (const directory of directories) {
+    const directoryPath = path.join(root, directory);
+
+    const instanceYml = loadInstanceConfig(directoryPath);
+
+    if (instanceYml && instanceYml.component) {
+      componentDirectoryFound = true;
+    }
   }
-  return false;
+
+  return componentDirectoryFound;
 };
 
 const getOutputs = (allComponentsWithOutputs) => {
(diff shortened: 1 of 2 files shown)


eahefnawy:
We don't have to call it a components template directory. I guess it's a generic term. But in your example, running `sls login` in this "Projects" directory would fire up the components experience because there are components in sub directories, and that's the intended behavior.

The framework login experience is already the default experience, here we are just prioritizing components in this particular case because it's relevant for components more than the framework. I wouldn't call that greedy 😄

medikoo:
> The framework login experience is already the default experience, here we are just prioritizing components in this particular case because it's relevant for components more than the framework. I wouldn't call that greedy 😄

Ok, but wouldn't that kill e.g. `sls create`, AWS interactive CLI and other commands  which work outside of service context in that case?

If I understand correctly after we merge that in, I will not be able to create a v1 project via `sls create` when being in _Projects_ folder.

Same way interactive CLI will stop working in it's context (which also is used to create v1 projects from templates).

To me it looks as quite breaking change, also I believe for many it may appear as unwanted behavior

eahefnawy:
@medikoo hmmm those are valid concerns. 🤔 

**Here's our motivation:** We want users who downloaded a component's template (ie. Full Stack App) to login via the Components flow, could you think of better ways to accomplish this?

medikoo:
> Here's our motivation: We want users who downloaded a component's template (ie. Full Stack App) to login via the Components flow, could you think of better ways to accomplish this?

Where is the source code of _Full Stack App_ template? (I remember they were here in `templates` folder, but I don't see a _full stack_ template there now)

eahefnawy:
@medikoo right here:
https://github.com/serverless-components/fullstack-app

eahefnawy:
@medikoo I've made the updates as discussed on Slack and tested these use cases. Will merge so that @ac360 is unblocked for an upcoming stream demo.

src/cli/utils.js
```diff
@@ -439,11 +439,31 @@ const isProjectPath = async (inputPath) => {
...
+
+  for (const directory of directories) {
+    const directoryPath = path.join(root, directory);
+
+    const instanceYml = loadInstanceConfig(directoryPath);
+
+    // if no yaml file found, or not a component yaml file
+    // then it's not a template directory
+    if (!instanceYml || !instanceYml.component) {
+      isTemplateDirectory = false;
```
medikoo, line 457: It'll be optimal to abort at this point, and do not investigate any further directories

Issue #620, opened 2020-05-04
Unexpected 'core-js@2' dependency


medikoo:
Regression introduced with https://github.com/serverless/components/pull/608

There are various issues with `core-js`.
- It list an advertisment on install. So currently when someone installs `npm install -g serverless` it sees: ![Screenshot 2020-05-04 at 12 35 10](https://user-images.githubusercontent.com/122434/80957601-c5a21000-8e03-11ea-82bd-4981463714fc.png)
- There are other issues related to `postinstall` script: https://github.com/serverless/serverless/issues/7654
- It weights 7MB, which makes it 6th largest dependency of Serverless Framework, and that significantly contributes to general size of a project and its standalone binary

hkbarton:
will look into this ASAP

hkbarton:
the issue is now fixed

PR #622, opened 2020-05-05
fix: remove core-js dependency for addressing issues-620

hkbarton:
## What has been implemented?

This PR addressed https://github.com/serverless/components/issues/620


src/cli/index.js
```diff
@@ -38,11 +42,16 @@ module.exports = async () => {
 
   // set global proxy agent if it's configured in environment variable
   if (process.env.HTTP_PROXY || process.env.HTTPS_PROXY) {
-    // set GLOBAL_AGENT_ENVIRONMENT_VARIABLE_NAMESPACE to empty, so the
-    // global-agent will use HTTP_PROXY/HTTPS_PROXY environment variables
-    // without GLOBAL_AGENT_ prefix.
-    process.env.GLOBAL_AGENT_ENVIRONMENT_VARIABLE_NAMESPACE = '';
-    proxyBootstrap();
+    if (semver.gte(process.version, 'v11.7.0')) {
```
medikoo, line 45: Are you sure no lower versions are supported?

By `engine` field, it appears there's support for Node.js v6+: https://github.com/TooTallNate/node-https-proxy-agent/blob/5.0.0/package.json#L54

There was once issue with Node.js v10 support, but it looks as addressed: https://github.com/TooTallNate/node-https-proxy-agent/issues/88

(still I'm just referring to what I see in repo, I haven't tested any Proxy setup)

hkbarton: It's the overriding globalAgent on node `http` and `https`. Overriding this property is only supported starting from node `11.7.0` (https://nodejs.org/uk/blog/release/v11.7.0/). Without being able to override globalAgent, we have to either config agent on every node HTTP client calls or monkey patch the method like `http.get` which I think it's not an ideal way to do. Thus, I decided only support proxy setting on node version 11.7 or above.

medikoo: 👍

PR #625, opened 2020-05-07
fix: should allow reference variables in env file at app level

hkbarton:
## What has been implemented?

This change allows reference variables in `.env` at the top level of `serverless.yml`, before referencing variables is only allowed in the `inputs` section.

### DIFF
--- a/src/cli/commands-cn/run.js
+++ b/src/cli/commands-cn/run.js
@@ -52,12 +52,6 @@ module.exports = async (config, cli, command) => {
   options.debug = config.debug;
   options.dev = config.dev;
 
-  // Connect to Serverless Platform Events, if in debug mode
-  if (options.debug) {
-    // TODO: to be implement for tencent
-    delete options.debug;
-  }
-
   if (command === 'deploy') {
     // Warn about dev agent
     if (options.dev) {
--- a/src/cli/commands-cn/utils.js
+++ b/src/cli/commands-cn/utils.js
@@ -8,7 +8,7 @@ const path = require('path');
 const fs = require('fs');
 const args = require('minimist')(process.argv.slice(2));
 const { utils: platformUtils } = require('@serverless/platform-client-china');
-const { loadInstanceConfig, resolveInputVariables } = require('../utils');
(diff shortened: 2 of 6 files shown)


src/cli/commands-cn/utils.js
```diff
@@ -88,7 +88,7 @@ const loadTencentInstanceConfig = async (directoryPath) => {
   if (instanceFile.inputs) {
     // load credentials to process .env files before resolving env variables
     await loadInstanceCredentials(instanceFile.stage);
-    instanceFile.inputs = resolveInputVariables(instanceFile.inputs);
+    instanceFile = resolveInputVariables(instanceFile);
```
medikoo, line 91: It'll be good to rename the util to `resolveVariables`, current name suggests it's strictly for _inputs_ and that makes this change confusing

src/cli/commands-cn/run.js
```diff
@@ -55,7 +55,6 @@ module.exports = async (config, cli, command) => {
   // Connect to Serverless Platform Events, if in debug mode
   if (options.debug) {
     // TODO: to be implement for tencent
-    delete options.debug;
   }
```
medikoo, line 58: I believe whole condition should be removed (I guess also it won't pass lint)

src/cli/commands-cn/run.js
```diff
@@ -52,12 +52,6 @@ module.exports = async (config, cli, command) => {
   options.debug = config.debug;
   options.dev = config.dev;
 
-  // Connect to Serverless Platform Events, if in debug mode
-  if (options.debug) {
-    // TODO: to be implement for tencent
-    delete options.debug;
-  }
-
```
medikoo, line 60: Following https://github.com/serverlessinc/platform-cn/pull/10#discussion_r421274798

I think it'll be breaking for currently published version of `@serverless/platform-client-china`.

I think first we need to publush v1.0.12 of @serverless/platform-client-china and bump `package.json` again here

hkbarton: that's right, let me do that first.

PR #628, opened 2020-05-07
Fix reloading yaml file in dev mode

eahefnawy:
Memoizing was getting in the way of reloading the yaml file with the same args on file change.

### DIFF
--- a/src/cli/commands/utils.js
+++ b/src/cli/commands/utils.js
@@ -20,7 +20,12 @@ const { join, basename } = require('path');
 
 const { readFileSync } = require('fs');
 const ini = require('ini');
-const { fileExistsSync, loadInstanceConfig, resolveVariables } = require('../utils');
+const {
+  fileExistsSync,
+  loadInstanceConfig,
+  loadInstanceConfigUncached,
+  resolveVariables,
+} = require('../utils');
 
 const getDefaultOrgName = async () => {
   const res = readConfigFile();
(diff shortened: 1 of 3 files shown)


eahefnawy:
@medikoo that makes sense 👍

eahefnawy:
@medikoo Updated. Could you take another look?

src/cli/commands/utils.js
```diff
@@ -157,8 +162,13 @@ const loadInstanceCredentials = () => {
  * Reads a serverless instance config file in a given directory path
  * @param {*} directoryPath
  */
-const loadVendorInstanceConfig = async (directoryPath) => {
-  let instanceFile = loadInstanceConfig(directoryPath);
+const loadVendorInstanceConfig = async (directoryPath, cached = true) => {
```
medikoo, line 165: It's nicer to use options for such data (it highly improves readablity, and provides room for other optional arguments in a future, see also https://ariya.io/2011/08/hall-of-api-shame-boolean-trap).

Additionally better design in case of booleans is to have `false` as default (then we can simply confirm truthiness of value as `options.booleanOption`. In case of `false` override, solid check would be: `!options.booleanOption && options.booleanOption != null` which is way more verbose)

Therefore I would change it to `{ disableCache: false }`

eahefnawy: done

src/cli/commands/utils.js
```diff
@@ -157,8 +162,13 @@ const loadInstanceCredentials = () => {
...
  */
-const loadVendorInstanceConfig = async (directoryPath) => {
-  let instanceFile = loadInstanceConfig(directoryPath);
+const loadVendorInstanceConfig = async (directoryPath, options = { disableCache: false }) => {
+  let instanceFile;
+  if (!options.disableCache && options.disableCache !== null) {
+    instanceFile = loadInstanceConfig(directoryPath);
+  } else {
+    instanceFile = loadInstanceConfigUncached(directoryPath);
+  }
```
medikoo, line 171: It should rather be as simple as:

```javascript
const instanceFile = options.disableCache
  ? loadInstanceConfigUncached(directoryPath)
  : loadInstanceConfig(directoryPath)
```


Note that the way you configured it `disableCache: null` will be assumed as `disableCache: true`.

(sorry maybe my original comment was confusing, I was referring to situation where we stand on `true` as default, and`false` as opt-in, and it's not the case now, and that's good)

eahefnawy: ah yeah, I thought you were referring to this one. No worries, I just updated it.

PR #642, opened 2020-05-22
Separate concerns

ac360:
## What has been implemented?

A recent change to the experience for China-based users broke the experience for all users.

At the heart of this problem is the fact that our architecture boots up both of these experiences, when only one is necessary.  Meanwhile, both of these experiences are not testing eachother, which results in breaking the other.

So, let's stop loading both of these experiences.  Performance will also improve as a result of this.

This change is an initial step in that direction.  It extracts the China-based user direction from the rest of the China-based logic, and uses that to optionally load the China-focused experience.

Next, the global-focused experience should not be loaded for China-based users.

By separating these concerns, we reduce bugs, create autonomy and innovate faster!


tinafangkunding:
I got a concern about this:

Sometimes, when users want to use China-based CLI while their IP is outside China (in Japan/Singapore, etc.) Their default experience is AWS supported. Currently, we'll guide them by adding configure in `.env`
```
SERVERLESS_PLATFORM_VENDOR=tencent 
```
is this logic gets changed after the separation?

PR #645, opened 2020-05-28
feat: showing progress/speed when deploy large code folder

hkbarton:

### DIFF
--- a/src/cli/commands-cn/run.js
+++ b/src/cli/commands-cn/run.js
@@ -64,6 +64,13 @@ module.exports = async (config, cli, command) => {
 
     // run deploy
     cli.status('Deploying', null, 'white');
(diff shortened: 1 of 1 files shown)


ac360:
I saw a demo of this.  The feature is cool, but the CLI output doesn't have great presentation imo.  CLI design is key for good DX.

Let's do this:

* If it's uploading, say "Uploading" instead of "Deploying".
* If it's uploading, let's put the stats data next to the word uploading and get rid of the brackets.
* If the data (e.g. percentage) is not accurate, cut it until it works.  There should never be a broken experience in front of customers.  We are what we ship.

Example:

```
3s > express-api > Uploading 10% 143/KB second ...
```

hkbarton:
updated!

PR #646, opened 2020-05-29
chore: output dashboard url after successful deployment

hkbarton:


eahefnawy:
Looks good! The `info` command also has a dashboard log if I remember correctly

hkbarton:
updated info as well

Issue #640, opened 2020-05-19
Feature Proposal: support inputs for custom component method


hkbarton:
Currently, we only support inputs for the `deploy` method on the component. We do support using `sls custom-method` to run any method in the component, but there is no way to pass in any inputs to those custom methods. I'm proposing adding a new section in `serverless.yml` to define the inputs for the custom method:

```
component: scf
name: demo
app: demoApp
stage: dev

inputs: # this is only for deploy
   src: ./src

actionInputs: # this section defines any inputs for arbitrary method in the component
  invoke: # suppose we have a method called invoke in scf component
    param1: "test1" # invoke method in scf component can read this parameter from inputs
    src: ./test.json # action inputs also supports uploading some files and getting those files in custom method
  test: 
    src: ./test-case.json
```

eahefnawy:
Or maybe:

```yml
inputs:
  deploy:
    src: ./src

  invoke:
    test: abc
```

I feel like the structure is getting more and more verbose in both approaches though. I would love to keep the yaml file simple. @hkbarton what is the custom action you wanna pass inputs to?

hkbarton:
@eahefnawy for a specific use case, I think the cloud function invoke is a useful one. so the user can use `sls invoke` to directly access and run a cloud function with some test input. This is useful for test cloud function without fire the real trigger event.

medikoo:
> Or maybe:
> 
> ```yml
> inputs:
>   deploy:
>    src: ./src
> 
>  invoke:
>    test: abc
> ```

I think it cannot work reliably together with maintaining support for current way. As otherwise what would be the logic to detect if user configured inputs for different commands or just for deploy command. There seems no clean and safe way to resolve that.

--- 

Maybe better would be to introduce something as:

```yaml
commandInputs:
  deploy:
    src: ./src

  invoke:
    test: abc
```

And at same time maintain support for:

```yaml
inputs: # deploy only inputs
  src: ./src
```

ac360:
This is a big decision and we should get it right.

Just curious, what are some real-world use-cases that would drive this (e.g. SCF "invoke") and what are some potential alternatives (e.g. outside of YAML)?

hkbarton:
[...]

PR #651, opened 2020-06-04
feat: support inputs for custom commands

hkbarton:
## What has been implemented?

Closes https://github.com/serverless/components/issues/640

### DIFF
--- a/src/cli/commands-cn/utils.js
+++ b/src/cli/commands-cn/utils.js
@@ -9,6 +9,7 @@ const fs = require('fs');
 const args = require('minimist')(process.argv.slice(2));
 const { utils: platformUtils } = require('@serverless/platform-client-china');
 const { loadInstanceConfig, resolveVariables } = require('../utils');
+const { mergeDeepRight } = require('ramda');
 
 const updateEnvFile = (envs) => {
   // write env file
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -37,6 +37,7 @@
     "axios": "^0.19.2",
     "chalk": "^2.4.2",
     "chokidar": "^3.3.1",
+    "deepmerge": "^4.2.2",
```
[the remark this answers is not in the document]
hkbarton: done

src/cli/commands-cn/utils.js
```diff
@@ -212,11 +213,24 @@ const getInstanceDashboardUrl = (instanceYaml) => {
...
+const setInputsForCommand = (instanceYaml, command, config) => {
+  if (instanceYaml.commandInputs) {
+    const defaultInputs = command === 'deploy' ? instanceYaml.inputs : {};
+    instanceYaml.inputs = instanceYaml.commandInputs[command] || defaultInputs;
+    // merging inputs from command args, e.g. slcc deploy --inputs.src="./new-src"
+    // will be merged into inputs.src
+    if (config.inputs) {
+      instanceYaml.inputs = mergeDeepRight(instanceYaml.inputs, config.inputs);
+    }
+  }
```
medikoo, line 225: I think to not break things, it should be more as:

```javascript
if (instanceYaml.commandInputs) {
  // .. logic as above
} else if (command !== 'deploy') {
  instanceYaml.inputs = {}
}
```

Otherwise we will start leaking `inputs` to other commands than `'deploy'`

hkbarton: good catch!

hkbarton:
> Looks good, but I thought we want this feature in AWS as well, no?

cc @ac360 if we want this in the global version I can work on the PR for that.

PR #657, opened 2020-06-10
fix: fix template downloading doesn't work with proxy

hkbarton:
Fix https://github.com/serverless/components/issues/634


medikoo:
Good one @hkbarton !

PR #661, opened 2020-06-12
Support backend notifications

medikoo:
Brings support for backend notifications in Components v2 (as outlined in https://github.com/serverless/roadmap/issues/321)

- [x] ~Depends on https://github.com/serverless/utils/pull/53 which needs to be merged first and `@serverless/util` at `v1.0.0` needs to be published~

### DIFF
--- a/src/cli/notifications/request.js
+++ b/src/cli/notifications/request.js
@@ -1,0 +1,53 @@
+'use strict';
+
+const notificationsUrl = require('@serverless/utils/analytics-and-notfications-url');
+
+module.exports = async (payload) => {
+  if (!notificationsUrl) return null;
+
+  const http = notificationsUrl.startsWith('http:') ? require('http') : require('https');
+
+  const requestBody = JSON.stringify(payload);
+
+  return new Promise((resolve) => {
+    const request = http.request(
+      notificationsUrl,
+      {
+        method: 'POST',
+        headers: {
+          'Content-Type': 'application/json',
+          'Content-Length': requestBody.length,
+        },
+      },
+      (response) => {
+        if (response.statusCode !== 200) {
+          // ignore errors
+          resolve(null);
+          return;
+        }
+        let responseBody = '';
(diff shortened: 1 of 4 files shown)


src/cli/notifications/request.js
```diff
@@ -0,0 +1,53 @@
...
+
+module.exports = async (payload) => {
+  if (!notificationsUrl) return null;
+
+  const http = notificationsUrl.startsWith('http:') ? require('http') : require('https');
+
+  const requestBody = JSON.stringify(payload);
+
+  return new Promise((resolve) => {
+    const request = http.request(
```
hkbarton, line 13: should we consider putting this into common utils(the parent level util)? so we can leverage this in other places, I think I already made similar use of node HTTP module in some previous PRs.

medikoo: Ideally I believe we should use some _request_ library that's chosen for this project.

Relying on bare `http` wrapper, may in turn force us to build step by step our own library which is not great direction I think.

I see that `axios` was used, and I was relying on it previously, but then it appeared that it doesn't support `HTTP_PROXY` and it's the reason I decided to rely on plain _request_ for a meantime (as you did with that fix: (https://github.com/serverless/components/pull/657)

@eahefnawy what do you suggest in given situation?

eahefnawy: I'd love to have a single library for everything, but if the CN version has certain requirements, we really don't want to complicate the global version as a result. That's why we're keeping them separate.

medikoo: @eahefnawy Note we're speaking here about generic HTTP handler. HTTP works same in US and China (fact that proxy related issue affected some Chinese users, doesn't mean it won't affect some users in other parts of the world)

Also this part of code under which we comment now is platform generic (would be run for both US and China).

Ideally `request` that's in utils should be refactored to not use Axios, but solution that ensures HTTP_PROXY is supported, then here we could rely on it here. Still I think it's a task for a different PR.

PR #663, opened 2020-06-16
Fix CLI hanging on error

eahefnawy:
Fixing regression caused by #648 and follow up for #658

Wrapping `run` and `runAll` code in try/catch to be able to disconnect the sdk.

### DIFF
--- a/src/cli/commands/run.js
+++ b/src/cli/commands/run.js
@@ -132,14 +132,9 @@ module.exports = async (config, cli, command) => {
       cli.logOutputs(instance.outputs);
     }
 
-    sdk.disconnect();
     cli.close('success', 'Success');
-  } catch (e) {
-    if (sdk) {
-      sdk.disconnect();
-    }
-
-    throw e;
+  } finally {
+    sdk.disconnect();
   }
   return null;
 };
--- a/src/cli/commands/run.js
+++ b/src/cli/commands/run.js
@@ -14,123 +14,132 @@ const {
 const runAll = require('./runAll');
 
 module.exports = async (config, cli, command) => {
-  if (config.all) {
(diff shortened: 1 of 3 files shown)


src/cli/commands/run.js
```diff
@@ -14,123 +14,132 @@ const {
...
+    }
 
-  sdk.disconnect();
-  cli.close('success', 'Success');
+    sdk.disconnect();
+    cli.close('success', 'Success');
+  } catch (e) {
+    if (sdk) {
+      sdk.disconnect();
+    }
```
medikoo, line 140: Wouldn't it be cleaner to wrap just part that's after `const sdk = ` with `try/finally` as:

```javascript
... // pre `const sdk = ` logic
const sdk = ...
try {
  ... // post `const sdk =` logic
} finally {
  sdk.disconnect();
}
```
```

Comment (eahefnawy) [reply src/cli/commands/run.js:140]: Good one 👍

File: src/cli/commands/run.js [hunk 2]
```diff
@@ -14,123 +14,127 @@ const {
...
-
-  sdk.disconnect();
-  cli.close('success', 'Success');
+      cli.log();
+      cli.logOutputs(instance.outputs);
+    }
 
+    cli.close('success', 'Success');
+  } finally {
+    sdk.disconnect();
```
medikoo, line 137: It'll crash if error happens before `sdk` is defined.

In comment I proposed to leave `sdk` to be defined as _constant_  (via `const`) - so no changes to this line, and start `try` wrapper right after `sdk` is initialized. Then in finally clause we're 100% certain `sdk` was initialized.

eahefnawy: done

PR #664, opened 2020-06-17
fix: can't quit dev mode

hkbarton:


src/cli/commands-cn/dev.js
```diff
@@ -115,6 +116,7 @@ module.exports = async (config, cli) => {
     const deployedInstance = await deploy(sdk, instanceYaml, instanceCredentials);
     if (await updateDeploymentStatus(cli, deployedInstance, false)) {
       cli.close('success', 'Dev Mode Closed');
+      process.exit(0);
```
medikoo, line 119: @hkbarton please see: https://github.com/serverless/components/pull/648

Can you revert and propose solution  that does not introduce a hard process exit?

PR #665, opened 2020-06-17
chore: make dev mode exiting more robust

hkbarton:


medikoo:
@hkbarton this introduced a regression against https://github.com/serverless/components/pull/665 can we revert?

PR #670, opened 2020-06-17
Adds support for templates registry

eahefnawy:


src/cli/utils.js
```diff
@@ -206,6 +206,54 @@ const loadComponentConfig = (directoryPath) => {
   return componentFile;
 };
 
+/**
+ * Reads a serverless config file in a given directory path
+ * @param {*} directoryPath
+ */
+const loadServerlessFile = (directoryPath) => {
```
astuyve, line 213: I have a similar method name and utility class needed to read and write the serverless file for onboarding. Think we can consolidate?
https://github.com/serverless/components/compare/add-init-tokens-to-components-cli#diff-9e3c0d734a00c6ce24205b3934539a18R1

eahefnawy: Sure. Let's just keep in mind that we also support `serverless.json`. I've also added error handling for v1 cases where the user could be using yaml syntax specific to CloudFormation, which caused the CLI to crash.

src/legacy.js
```diff
@@ -13,7 +13,7 @@ const runningComponents = () => {
   let instanceConfig;
 
   // load components if user runs "sls registry" or "sls --all" or "sls --target" (that last one for china)
-  if (process.argv[2] === 'registry' || args.all || args.target) {
+  if (process.argv[2] === 'publish' || process.argv[2] === 'registry' || args.all || args.target) {
```
astuyve, line 16: This is only getting larger, I moved to an allowList pattern here, since I'll need it for init: https://github.com/serverless/components/compare/add-init-tokens-to-components-cli#diff-35e12ed3c51395fa98e153f4b59c76a4R10

hkbarton: agree, we should put this in a list

PR #676, opened 2020-06-22
feat: support registry template

hkbarton:
## What has been implemented?

* Template registry for CN
* backward compatibility for component registry

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
     "figures": "^3.2.0",
     "fs-extra": "^8.1.0",
     "globby": "^10.0.2",
+    "got": "^9.6.0",
     "graphlib": "^2.1.8",
     "https-proxy-agent": "^5.0.0",
     "ini": "^1.3.5",
--- a/src/cli/commands-cn/create.js
+++ b/src/cli/commands-cn/create.js
@@ -4,8 +4,12 @@
  * CLI: Command: CREATE
  */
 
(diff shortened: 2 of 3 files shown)


src/cli/commands-cn/create.js
```diff
@@ -15,23 +19,43 @@ module.exports = async (config, cli) => {
...
+        path: url.path,
+        method: 'GET',
+      },
+      (res) => {
+        const stream = res.pipe(fs.createWriteStream(tmpFilename));
+        stream.on('error', reject);
+        stream.on('finish', resolve);
+      }
+    );
+    req.end();
```
medikoo, line 52: I think we also need to observe `error` on request (as e.g. got does here: https://github.com/sindresorhus/got/blob/56df87671c741251a657fd57f60857ed9927f76b/source/core/index.ts#L1253) as otherwise such errors will thrown as uncaught exceptions.

I think to be on safe side we definitely should use some solid request library. Have you tried if `got` will work with `HTTP_PROXY` env vars?

package.json
```diff
@@ -42,6 +42,7 @@
     "figures": "^3.2.0",
     "fs-extra": "^8.1.0",
     "globby": "^10.0.2",
+    "got": "^11.3.0",
```
medikoo, line 45: We need to maintain support for Node.js v8, hence we should use `got` v9 (that would be validated with CI/CD if it would be already merged)

PR #678, opened 2020-06-22
Feature: sls init

astuyve:
## What has been implemented?
Implements a new command, `sls init`, which initializes new projects from one of two sources:
- package name, from the registry
- initToken, from the SF Pro Backend

## Steps to verify

- [ ] Link this feature branch using `npm link` or `npm-cross-link`
- [ ] run `SERVERLESS_PLATFORM_STAGE=dev sls init fullstack-app`
- [ ] It should pull down the app and configure it for you.
- [ ] Have me create a token for you in the backend
- [ ] run `SERVERLESS_PLATFORM_STAGE=dev sls init <token>`
- [ ] It should pull down the app and configure it for you, setting your org and app name information as well.

- [ ] Run `SERVERLESS_PLATFORM_STAEG=dev sls init fooobar`
- [ ] You should get an error, that no such package exists.


ac360:
Testing it out now.

Ran into this error when trying `init`

Screen Shot 2020-06-23 at 3 54 56 PM

ac360:
# Error 1

* Logged in
* I'm pretty sure my DEV stage config is set correctly and it cannot find the `fullstack-app` template.
* Additionally, error catching/handling seems broken.  It looks like 2 errors are being reported:

```
init-tests: components init fullstack-app

 Error: Can't find template: fullstack-app, run 'sls registry' to see available templates.
    at CLI.error (/Users/ac/projects/serverless/code/components-core/src/cli/CLI.js:298:15)
    at Object.run (/Users/ac/projects/serverless/code/components-core/src/cli/commands/initializer/index.js:43:15)
    at process._tickCallback (internal/process/next_tick.js:68:7)

Serverless › Can't find template: fullstack-app, run 'sls registry' to see available templates.


 Error: Something went wrong, please try again
    at CLI.error (/Users/ac/projects/serverless/code/components-core/src/cli/CLI.js:298:15)
    at Object.module.exports [as init] (/Users/ac/projects/serverless/code/components-core/src/cli/commands/init.js:23:9)
    at process._tickCallback (internal/process/next_tick.js:68:7)

Serverless › Something went wrong, please try again

init-tests:
```

ac360:
@astuyve If we are going to start separating command logic into multiple files, let's find a cleaner organization for that.  Keep in mind, messy codebases can cause more pain than normal because the China effort is emulating a lot of what we're doing, so a 2x mess will result.

How about we still with the same names and just create a folder in the `commands` folder that matches the name of the command:

```
init/
init.js
```

This puts a nice pattern in place that seems like it can scale well.

ac360:
Current status:

I tested without a token, just using the registry plain. 

I ran into many unhelpful errors and usability issues.  I've submitted fixes for those w/ clear code commits.

I was able to publish the fullstack-app template and install it.  The process is great!

However, I have not tested w/ init tokens.  

Also, the last part is broken in a few ways:  

* The CLI tells you to `serverless dev`.  However, the `fullstack-app` will not work with that command, nor will any template that has multiple apps.  This command will also not work w/ non-code related components (e.g. dynamodb).   This command should be avoided for components based projects and `serverless deploy` should be used instead.

* `serverless deploy` is not currently supported and only `serverless deploy --all` can be used.  However, `serverless deploy --all` is now broken because it's trying to deploy the `serverless.yml` at the parent, which is just used to declare a template.  This should NOT be packaged into the resulting template cc @eahefnawy.

* Also, I now run into a `Error: Missing "org" property in serverless.yml` when trying to run `sls deploy -all`

* When

src/cli/serverlessFile.js
```diff
@@ -0,0 +1,121 @@
...
+    const ymlObject = await loadServerlessFile(servicePath);
+    if (ymlObject) {
+      if (orgName) ymlObject.org = orgName;
+      if (appName) ymlObject.app = appName;
+      if (serviceName) ymlObject.service = serviceName;
+
+      await writeServerlessFile(cli, servicePath, ymlObject);
+    }
+  }
+};
```
medikoo, line 114: It'll be nice to reuse same logic we have here: https://github.com/serverless/enterprise-plugin/blob/a0a1614e3eb2ab22df925ff2158f6d315cfdd829/lib/interactiveCli/writeOrgAndApp.js which could be moved to `@serverless/utils`

src/cli/utils.js
```diff
@@ -675,14 +627,13 @@ module.exports = {
...
   fileExistsSync,
-  readFileSync,
+  readAndParseSync,
   isYamlPath,
   isJsonPath,
   resolveVariables,
   getDirSize,
   pack,
   getInstanceDashboardUrl,
-  loadServerlessFile,
```
medikoo, line 685: All modules that depend on it should be updated to use new location.

As this package has no tests, this issue is not picked by CI :/

src/cli/utils.js
```diff
@@ -675,14 +627,13 @@ module.exports = {
   request,
   fileExists,
   fileExistsSync,
-  readFileSync,
+  readAndParseSync,
```
medikoo, line 630: This probably can be removed from export (no modules use that util)

src/legacy.js
```diff
@@ -6,20 +6,17 @@ const {
   utils: { isChinaUser },
 } = require('@serverless/platform-client-china');
 
+// These keywords should route to components CLI, not sls cli.
+const componentKeywords = ['registry', 'init', 'publish'];
```
medikoo, line 10: nit: It'll be neater to use _set_ instead of _array_

eahefnawy:
Thanks @astuyve ... I took it for a spin, and I generally agree with @ac360, along with minor styling suggestions:

1. We should log the logo at the beginning of the command with `cli.logLogo()`
2. We try to keep the status message as short as possible, so I think it'd be better to log paths as a normal log instead of a status
3. It would be great if we could match the styling/colors to that of other commands, rather than making the status all grey.

These are minor, and honestly, the CLI still needs a lot of work to reduce duplication and put together some conventions now that we have an increasing number of commands. So this could be addressed later.

Other than that, I was able to deploy the `fullstack-app`:

```
fullstack-app (master) node ../../../components/bin/bin deploy --all

serverless ⚡ framework (dev)
Action: "deploy --all" - Stage: "dev" - Org: "eahefnawy" - App: "fullstack-app" - Name: "fullstack-app"

fullstack-db: 
  name:    fullstack-db-dev
  arn:     arn:aws:dynamodb:us-east-1:552750238299:table/fullstack-db-dev
  region:  us-east-1
  indexes: 
    gsi1: 
      name: gsi1
      arn:  arn:aws:dynamodb:us-east-1:552750238299:table/fullstack-db-dev/index/gsi1

fullstack-site: 
  bucket:    website-jppae3
  bucketUrl: http://website-jppae3.s3-website.us-east-1.amazonaws.com
  url:       https://d3se7v0zk815fk.cloudfront.net

fullstack-permissions: 
  name: fullstack-permissions-dev
  arn:  arn:aws:iam::552750238299:role/fullstack-permissions-dev

fullstack-api: 
  url: https://nz68loo0nd.execute-api.us-east-1.amazonaws.com

35s › fullstack-app › Success

fullstack-app (master) 
```

src/cli/commands/initializer/index.js
```diff
@@ -0,0 +1,76 @@
...
+   * @param {*} cliParams
+   */
+  async run(cli, cliParam) {
+    cli.start('Fetching app configuration');
+    let templateUrl;
+    let directory;
+    let serviceName;
+    let tenantName;
+
+    // packages and tokens are both strings. First see if the arg is a token
```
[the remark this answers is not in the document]
astuyve: Conversation happening in slack

src/cli/commands/initializer/unpacker.js
```diff
@@ -0,0 +1,56 @@
...
+    this.tenantName = tenantName;
+    this.serviceName = serviceName;
+  }
+
+  /**
+   * Recursive method
+   * @param {*} dir
+   */
+  async unpack(dir) {
+    process.chdir(dir);
```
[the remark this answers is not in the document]
astuyve: Agreed, will move it to the init logic before the handoff here.

src/cli/commands/initializer/unpacker.js
```diff
@@ -0,0 +1,56 @@
...
+    // Check if the directory contains a serverless.yml/yaml/json/js.
+    // If it does, we need to unpack it
+    if (getServerlessFilePath(dir)) {
+      this.cli.status(`Unpacking '${dir}'`);
+      if (fs.existsSync('package.json')) {
+        this.cli.status(`Installing node_modules via npm in ${dir}`);
+        await spawn('npm', ['install']);
+      }
+      if (fs.existsSync('yarn.lock')) {
```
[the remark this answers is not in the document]
astuyve: Because some people use yarn, and some people use npm.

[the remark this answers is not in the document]
astuyve: We won't know deterministically what the template author intended here, so we check for both.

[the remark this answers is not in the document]
medikoo: > Because some people use yarn, and some people use npm.

There are few more options, e.g. `pnpm`, why not supporting them as well? 

For simplicity I would stick just to npm (that's de-facto standard, rest comes and goes), and only if we have a very good reason (there's strong demand) consider other package managers.

Also doing two installs is quite bad, there should be just one.

src/cli/serverlessFile.js
astuyve: I had done that initially, back when this was implemented in the v1 cli.

In a different PR, I will refactor the logic from this PR into `@serverless/utils` and replace `writeOrgAndApp` in the plugin.

src/cli/utils.js
astuyve: ? I use `readAndParseSync` in this PR in several places.

src/cli/utils.js
astuyve: Done

src/cli/utils.js
medikoo: ah.. ok, sorry I didn't notice. I was referring to fact that in current `master`, `readFileSync` is not used by any module

src/cli/commands/init/index.js
```diff
@@ -0,0 +1,98 @@
...
+    try {
+      ({ templateUrl, directory, serviceName, tenantName } = await initTokenHandler(sdk, cliParam));
+    } catch (error) {
+      // Code doesn't exist
+      if (error.response && error.response.status === 404) {
+        cli.close(
+          'error',
+          `App token '${cliParam}' doesn't exist
+          \nGo to https://app.serverless.com to generate a new app token`
+        );
```
medikoo, line 43: I think we should `return` here (?)

src/cli/commands/init/index.js
```diff
@@ -0,0 +1,98 @@
...
+      }
```
medikoo, line 44: Shouldn't we throw error that doesn't fit condition?

src/cli/commands/init/index.js
```diff
@@ -0,0 +1,98 @@
...
+    // Set org attr in sls.yml for each
+    await unpacker.unpack(servicePath);
+  }
+  return directory;
+};
+
+const init = async (config, cli) => {
+  const maybeToken = config.params[0];
+  if (!maybeToken) {
+    cli.error('init command requires either a token or template URL');
```
medikoo, line 86: Shouldn't we return here?

src/cli/commands/init/unpacker.js
```diff
@@ -0,0 +1,53 @@
...
+        await spawn('npm', ['install'], { cwd: dir });
+      }
+
+      // Writes the tenantName and serviceName to the serverless.y(a)ml file
+      await writeMainAttrs(this.cli, dir, this.tenantName, this.serviceName);
+      const files = await fs.readdir(dir);
+      const result = await Promise.all(
+        files.map(async (file) => {
+          // Check if the file is a directory, or a file
+          const stat = await fs.stat(`${dir}/${file}`);
```
medikoo, line 40: nit: object returned by `fs.stat` is usually called _stats_ (not _stat_)

src/cli/commands/init/unpacker.js
```diff
@@ -0,0 +1,53 @@
...
+          if (stat.isDirectory()) {
+            return this.unpack(path.resolve(dir, file));
+          }
+          return true;
+        })
+      );
+      return result.filter(Boolean);
+    }
+    return true;
```
medikoo, line 49: What are the expected return values of this function?

I'm a bit confused, as I think it either returns `true` or array of `true` values. Wonder whether there's any intention behind it?

If its not expected to return anything meaningful, let's use plain `return` or `return null` (if lint complains). That'll make it clear.

src/cli/commands/init/utils.js
```diff
@@ -0,0 +1,26 @@
+'use strict';
+const { promisify } = require('util');
+const stream = require('stream');
+const fs = require('fs-extra');
+const got = require('got');
+const path = require('path');
+
+const pipeline = promisify(stream.pipeline);
```
medikoo, line 8: `stream.pipeline` is not supported in Node.js v8.

So we should use https://www.npmjs.com/package/stream.pipeline-shim instead, or consider releasing this under new major (with that we should also update support check in CI/CD from v8 to v10)

(Unfortunately such errors sneak not picked by CI/CD due to lack of tests)

src/cli/commands/init/unpacker.js
astuyve: Was true because linter freaked out, will use null

src/cli/commands/init/index.js
astuyve: yes

src/cli/commands/init/utils.js
astuyve: I copied this from the components-cn `init.js` command, and I thought components only supported v8 and newer?

medikoo: Hmm... so issue was already introduced in `components-cn` (cc @hkbarton )

>  I thought components only supported v8 and newer?

Yes, that's the case, but `stream.pipline` was introduced in v10

astuyve: I will use pipeline shim then

medikoo: @hkbarton just fixed in `commands-cn` -> https://github.com/serverless/components/pull/682

astuyve: Yeah, I had fixed it here too, but merged his changes in.

src/cli/commands/init/unpacker.js
```diff
@@ -0,0 +1,53 @@
...
+        files.map(async (file) => {
+          // Check if the file is a directory, or a file
+          const stats = await fs.stat(`${dir}/${file}`);
+          if (stats.isDirectory()) {
+            return this.unpack(path.resolve(dir, file));
+          }
+          return null;
+        })
+      );
+      return result.filter(Boolean);
```
medikoo, line 47: I guess for consistency we should return here `null` as well, but for real I think all `return` in scope of this function can be removed. This function is not intended to return anything, and there even no shortcuts here to end execution prematurely.

astuyve: Hmmm. I think this broke something. I removed it, and now I notice something.

Before this commit, we'd spend 30s every time running `npm install` in the `site` directory of the `fullstack-app`.

Now it seems to take the exact same 30s, but in a random subdirectory of the `fullstack-app`

ac360:
Fixed this issue with publishing components.

As of now, I tested:

* Publishing a Component.
* Publishing a Template.
* Installing a Template.

I couldn't finish testing due to errors.

**Existing Issues**

* I wanted to try the init flow but the dev environment authorization was broken as of Friday afternoon.  I was unable to register a new user.

* When downloading a template, all helpful comments are being stripped from `serverless.yml` hurting the new user experience greatly.  Quotes are also added which aren't necessary.  Whatever YAML parse we are using is likely doing this.

src/cli/commands/init/unpacker.js
medikoo: Do we have good understanding why it worked then, and it doesn't now?

astuyve: @medikoo I reverted to this change https://github.com/serverless/components/pull/678#discussion_r444788662 (with `result = await Promise.all`..., and that seemed to fix things.

medikoo: Do you understand why it fix things, or does it works by chance, and we have no idea why (?)

medikoo: As I looked into code, this method is invoked in two places.

1. https://github.com/serverless/components/blob/9e3a7ab18ec9f07493ae3d34f6f6085638dd465f/src/cli/commands/init/index.js#L81 What it returns is irrelevant
2. https://github.com/serverless/components/blob/9e3a7ab18ec9f07493ae3d34f6f6085638dd465f/src/cli/commands/init/unpacker.js#L42 - it's in self recursion, it's returned as result to `map` where again what it returns is irrelevant (_map_ here is just to collect _promises_, but what do they resolve with, doesn't play any role)

Now, which exactly version (commit hash?) doesn't work on your side?

astuyve: I think it's a red herring, I re-linked dependencies this morning and tried again, and it seems to work here. I'll let you know if I see it again

src/cli/utils.js
```diff
@@ -53,7 +51,7 @@ const request = async (options) => {
 
   let res;
   try {
-    res = await axios(requestOptions);
+    res = await got.post(options.endpoint, requestOptions);
```
medikoo, line 54: `response` as returned (or provided on _error_) by `axios` and `got` differs. `axios` uses [it's own format](https://github.com/axios/axios#response-schema) while got [returns Node.js response](https://github.com/sindresorhus/got/tree/v9.6.0#response), and API of both differs.

Therefore I believe this change break things, we need to alter processing of result _response_.

astuyve: It'd be great if we had literally any tests around this.

astuyve: @medikoo actually I think this entire method can be removed - I don't think we're using it anywhere at all: https://github.com/serverless/components/search?q=request&unscoped_q=request

medikoo: > It'd be great if we had literally any tests around this

Totally, it's definitey hard to move around safely without them

> @medikoo actually I think this entire method can be removed - I don't think we're using it anywhere at all: 

Hmm.. indeed. Let's remove it then (cc @eahefnawy)

PR #679, opened 2020-06-23
feat: support init for CN commands

hkbarton:
## What has been implemented?

* Change `create` to `init` for CN commands to match the new product spec. (`create` is never work for CN commands before, since all templates provided in this repo is for global version)
* bypass legacy CLI if `init` is given, similar to the change did in https://github.com/serverless/components/pull/678


src/cli/commands-cn/index.js
```diff
@@ -2,14 +2,14 @@
...
 const info = require('./info');
-const create = require('./create');
+const init = require('./init');
 const dev = require('./dev');
 const registry = require('./registry');
 
 module.exports = {
   run,
   info,
-  create,
```
medikoo, line 12: `create` was never used? Isn't that breaking change?

hkbarton: it never works for CN user, before the create just load global templates in this repo: https://github.com/serverless/components/tree/master/templates, so it actually causes more confuse for CN component user

PR #682, opened 2020-06-25
chore: support node v8 in init command

hkbarton:


package.json
```diff
@@ -54,9 +54,11 @@
     "minimist": "^1.2.5",
     "moment": "^2.27.0",
     "open": "^7.0.4",
+    "pify": "^5.0.0",
```
medikoo, line 57: @hkbarton I think we can use `require('util').promisify` instead (it's supported since Node.js v8)

PR #684, opened 2020-06-29
fix: save global http/https agent before changing them

hkbarton:


hkbarton:
> 👍
> 
> btw I think you need to setup your name and email in local git settings. Note that commits you submitted are not bound to your GitHub account.

ah, I switched to use a new computer, that's probably why, let me do that

PR #685, opened 2020-06-29
Remove --all flag and detect templates automatically 

eahefnawy:
## What has been implemented?

Remove --all flag and detect templates automatically. PR also fixes a minor bug when running remove (the "Success" message was logged twice)

With that in place, the Components CLI runs in the following cases:

- running `sls init`
- running `sls publish`
- running `sls registry`
- running inside a directory containing `serverless.yml` file of an instance
- running inside a directory containing child directories of instances and NO serverless.yml in the root
- NOT running `help` or `version`

## Steps to verify

- cd into a template directory
- instead of running `sls deploy --all` and `sls remove --all` as usual, just run `sls deploy` and `sls remove`
- the framework should detect that there are components in child directories and run them as a template.


src/legacy.js
```diff
@@ -12,12 +12,12 @@ const runningComponents = () => {
...
   let instanceConfig;
 
-  // load components if user runs "sls registry" or "sls --all" or "sls --target" (that last one for china)
+  // load components if user runs "sls registry" or "sls publish" or "sls --target" (that last one for china)
   if (
     process.argv[2] === 'publish' ||
     process.argv[2] === 'registry' ||
     process.argv[2] === 'init' ||
-    args.all ||
+    utils.runningTemplate(process.cwd()) ||
```
medikoo, line 20: This is quite problematic, as moving it here it means it'll be done before every v1 command, and check is far from optimal (it may imply an intensive file system traversal).

Ideally if it's done only if there's no `serverless.config.yml` and no `serverless.yml` file at working directory, and if there are no CLI args as `--help` or `--version`.

In general this design is problematic.  It'll be way better if we can state whether we deal with a _template_ or not, by some straightforward check (as it's done in all other ecosystem cases).

eahefnawy: Sounds good. I've pushed an update that skips the check if there's a config file in the root or help/version is running.

With that in place, if the cwd already has a component instance or a framework project, the template would be ignored completely and deployment would proceed as expected.

src/cli/utils.js
```diff
@@ -507,14 +507,27 @@ const isProjectPath = async (inputPath) => {
...
+  for (const configurationFile of possibleConfigurationFiles) {
+    if (fileExistsSync(path.join(inputPath, configurationFile))) {
+      return true;
+    }
+  }
+  return false;
+};
+
 const runningTemplate = (root) => {
+  if (isProjectPathSync(root)) {
```
medikoo, line 520: This logic should be included in below `try` clause

src/legacy.js
medikoo: Still, I believe we should put it into `  if (!componentConfig && !instanceConfig) {` condition. It'll be more optimal

src/cli/utils.js
eahefnawy: Ok

src/legacy.js
eahefnawy: hmm I'm not following ... doesn't fit the context above 🤔

eahefnawy: Also, that would require reading the file and running logic to determine if it's a component or an instance, how is that more optimal?

medikoo: It's less optimal for cases where we're in context of _component_, or _singular component instance_.

Note that in above cases, as proposed we always do three steps:
A. Check whether there's any configuration
B. Resolve eventual component config
C. Resolve eventual instance config

If this is put into `if (!componentConfig && !instanceConfig)` condition (as it is  in `master`), then only B & C checks will done in such case (and that's good enough)

For other cases it doesn't make difference

medikoo: > Also, that would require reading the file and running logic to determine if it's a component or an instance.

True, I missed part, that this would be less optimal for _multi component instance_ case, but check for it, is already drastically sub-optimal. I think in all cases this should be done as last resort (when we're sure we do not deal with any other option)

medikoo: Anyway if you like it more they way it is, it's indeed not a big deal, and difference is arguable.

I'm just concerned about this check, and automatically want to do anything so it's made in least possible cases :)

eahefnawy: Ok, I pushed an update that **I think** address your concern. Please take a look at the top of the `runningTemplates` method. It not only checks the existence of the yaml file, but it also makes sure that it is not a component instance nor a v1 framework. A component yaml file is irrelevant here because it's not "deployable" so it doesn't really conflict with the template anyway.

So now the template FS check does not happen at all if the cwd is a component instance or a v1 service, otherwise, we'll check for a template.

I've also limited this check to the `deploy` command, so that we don't run this on every invocation of the framework.

src/cli/utils.js
```diff
@@ -507,14 +505,34 @@ const isProjectPath = async (inputPath) => {
   return false;
 };
 
+const isDeployableProjectExists = (inputPath) => {
+  for (const configurationFile of possibleConfigurationFiles) {
+    const configurationFilePath = path.join(inputPath, configurationFile);
+    if (fileExistsSync(configurationFilePath)) {
+      const content = readFileSync(configurationFilePath);
```
medikoo, line 512: Why exactly we're parsing the file?  Isn't existence of `serverless.yml` in root folder good enough to state we're not in a _template_ project ?

Reading and parsing is way heavier than just confirming on file existence

eahefnawy: Ok I'm completely lost now. I thought that's what you wanted.

> Still, I believe we should put it into if (!componentConfig && !instanceConfig) { condition. It'll be more optimal

> Also, that would require reading the file and running logic to determine if it's a component or an instance, how is that more optimal compared to just checking for the existence of the file?

medikoo: No, I just opted to push it into same condition from which it was removed (specifically here: https://github.com/serverless/components/blob/56e01ca666e68949a72bc7c06bace7732ccee2ed/src/legacy.js#L50)_ :)

But after we discussed I agreed it doesn't make much difference (it slightly more optimal for root level component and instance configurations, but also less optimal for template configuration, and no difference in context of v1).

Check where you just checked for file existence was good (best we can do I believe).

Sorry for confusion, If you're still not sure we can further clarify on Slack

eahefnawy: Thinking about it a bit deeper, if you have a nested template, then add a `serverless.yml` at the root to publish it, then you wanna deploy it after. It should work. So just checking for the existence is probably not enough for this case

medikoo: Ok, in such case it'll be more optimal for components and non-template instances to have check put here https://github.com/serverless/components/blob/56e01ca666e68949a72bc7c06bace7732ccee2ed/src/legacy.js#L50 (as otherwise note that we're parsing components YAML twice).

But of course when putting there, we're enforce double parsing for template case.

Generally flow in proposed form is not as optimal as it can be, but probably acceptable.

src/legacy.js
```diff
@@ -9,16 +9,28 @@ const {
 const runningComponents = () => {
   const args = minimist(process.argv.slice(2));
 
+  const isRunningHelpOrVersion =
+    process.argv[2] === 'version' ||
+    process.argv[2] === 'help' ||
+    args.v ||
+    args.version ||
+    args.help;
```
medikoo, line 17: We should also have `args.h`

src/legacy.js
```diff
@@ -9,16 +9,28 @@ const {
 const runningComponents = () => {
   const args = minimist(process.argv.slice(2));
 
+  const isRunningHelpOrVersion =
+    process.argv[2] === 'version' ||
+    process.argv[2] === 'help' ||
```
medikoo, line 14: Do we plan to expose `help` or `version` via command and not via param (as e.g. `sls version`) in components?

eahefnawy: yeah they're in the above list. So running `sls version` in component project would just load the framework. When we want to have our own version/help commands in component, we can remove them.

src/legacy.js
eahefnawy: 👍

PR #692, opened 2020-07-01
fix: login command error

yugasun:
Fix below error:

image

Because we have do login here https://github.com/serverless/components/blob/master/src/cli/commands-cn/run.js#L24, so if command is login, do nothing here

### DIFF
--- a/src/cli/commands-cn/run.js
+++ b/src/cli/commands-cn/run.js
@@ -98,6 +98,9 @@ module.exports = async (config, cli, command) => {
     // run remove
     cli.status('Removing', null, 'white');
     await sdk.remove(instanceYaml, instanceCredentials, options);
+  } else if (command === 'login') {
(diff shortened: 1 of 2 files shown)


src/cli/commands-cn/run.js
```diff
@@ -98,6 +98,9 @@ module.exports = async (config, cli, command) => {
     // run remove
     cli.status('Removing', null, 'white');
     await sdk.remove(instanceYaml, instanceCredentials, options);
+  } else if (command === 'login') {
+    // we have do login upside, so if command is login, do nothing here
+    // no op
```
hkbarton, line 103: add cli.log("you'll be asked to scan QR code to login when you deploy if secrets is not set in env file")

yugasun: `utils.login()` method will print `Please scan QR code login from wechat`, so I think it's not necessary.

PR #695, opened 2020-07-01
Support removal of serverless.yml file at top of template

astuyve:
Modify recursive call to accept isTopLevel to ignore serverless.yml check before recursing

## What has been implemented?

Modifies recursive call to ignore a missing serverless.yml for the top-level use case.

## Steps to verify

- Run `sls init fullstack-app`
- note that node modules successfully installed

### DIFF
--- a/src/cli/commands/init/index.js
+++ b/src/cli/commands/init/index.js
@@ -103,7 +103,7 @@ const run = async (cli, cliParam) => {
     cli.status('Setting up your new app');
     // Recursively unpack each directory in a template
     // Set org attr in sls.yml for each
-    await unpacker.unpack(servicePath);
+    await unpacker.unpack(servicePath, true);
   }
   return directory;
 };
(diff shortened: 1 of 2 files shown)


src/cli/commands/init/index.js
```diff
@@ -103,7 +103,7 @@ const run = async (cli, cliParam) => {
     cli.status('Setting up your new app');
     // Recursively unpack each directory in a template
     // Set org attr in sls.yml for each
-    await unpacker.unpack(servicePath);
+    await unpacker.unpack(servicePath, true);
```
medikoo, line 106: It'll be way nicer to communicate it with `{ isTopLevel: true }`, otherwise it's not clear what this boolean is about

src/cli/commands/init/unpacker.js
```diff
@@ -22,10 +22,10 @@ class Unpacker {
    * Recursive method
    * @param {*} dir
    */
-  async unpack(dir) {
+  async unpack(dir, isTopLevel = false) {
```
medikoo, line 25: That could also be solved internally without changing `unpack` signature (recursive function could be internal)

PR #697, opened 2020-07-02
fix: fix deploy all for CN

hkbarton:


src/cli/commands-cn/run.js
```diff
@@ -15,7 +15,7 @@ const requestNotification = require('../notifications/request');
 const printNotification = require('../notifications/print-notification');
 
 module.exports = async (config, cli, command) => {
-  if (runningTemplate(process.cwd())) {
+  if (!config.target && runningTemplate(process.cwd())) {
```
medikoo, line 18: What does `config.target` state? Shouldn't we put same condition in AWS version?

src/cli/utils.js
```diff
@@ -446,11 +446,17 @@ const runningTemplate = (root) => {
...
       return false;
     }
-    return fse.readdirSync(root).every((fileOrDirName) => {
-      if (fileOrDirName.startsWith('.')) return true;
-      const fileOrDirPath = path.join(root, fileOrDirName);
-      if (!fse.statSync(fileOrDirPath).isDirectory()) return true;
-      const instanceYml = loadInstanceConfig(fileOrDirPath);
+    const subDirectories = fse
+      .readdirSync(root)
+      .filter((fileOrDirName) => fse.statSync(path.join(root, fileOrDirName)).isDirectory());
```
medikoo, line 451: This is not optimal, we should abort at first case which allows us to assume, that we do not deal with a template, and with logic as proposed we're retrieve _stats_ for each item in a folder (even if first already will disqualify the case)

src/cli/commands-cn/run.js
hkbarton: config.target is only CN, it used to deploy subproject separately, e.g. `sls deploy --target=./api`

src/cli/utils.js
hkbarton: hmm, not sure what did you mean? Could you explain more? What's the first case?

medikoo: 1. Non optimal (we're retrieving _stats_ for each folder item)

```javascript
const directories = fse.readdirSync(root).filter((...) => fse.statSync(...).isDirectory());

return directories.every(isComponent)
````

2. Optimal (we're retrieving _stats_ only when we still do not know the answer:

```javascript
return fse.readdirSync(root).every(f => {
  if (!fse.statSync(f).isDirectory())) return true;
  return isComponent(f);
})
````

File: src/cli/utils.js [hunk 3]
```diff
@@ -446,11 +446,17 @@ const runningTemplate = (root) => {
...
+
+    if (subDirectories.length === 0) {
+      return false;
+    }
+
+    return subDirectories.some((folderName) => {
```
medikoo, line 457: I've just noticed that this changes intended logic.

It should be `every` here. All folders should contain components, and not that _any_ should contain component

src/cli/utils.js
hkbarton: I see. I think my main change here is changing `every` to `some`, so I guess I need to do a regular for loop here instead of `some`, in order to optimize

medikoo: Question is, why do you think what's in master should be changed? What's not right about that logic?

hkbarton: The logic set assumption that in a template folder, every subfolder should include `serverless.yml`, otherwise deploy all will not work. I think this assumption is too strict, e.g. in `tencent-fullstack` template project, it includes a `script` folder which is a regular folder with some scripts inside, it doesn't include `serverless.yml` file

src/cli/utils.js
hkbarton: this is the exact logic which broke the deployment of Tencent fullstack template

src/cli/utils.js
medikoo: @hkbarton can you open discussion on Slack about that.. I think we really need to rethink on how we detect the template with @eahefnawy and @ac360

medikoo: I think also we should not have drop support for `--all` as that was kind of breaking (cc @eahefnawy )

hkbarton: `--all` is still supported I think

medikoo: It was removed with https://github.com/serverless/components/pull/685/ but it's just in `canary` so far. No regular version was published with it yet.

medikoo: @hkbarton in light of discussion on Slack, just change it so one `some` iteretion is made (no prior `filter` call), and we'd leave it for @eahefnawy for approval.

src/cli/utils.js
```diff
@@ -446,16 +446,18 @@ const runningTemplate = (root) => {
...
-      // then it's not a template directory
-      return instanceYml && instanceYml.component;
-    });
+    const itemNames = fse.readdirSync(root);
+
+    for (const itemName of itemNames) {
+      const itemPath = path.join(root, itemName);
+      if (fse.statSync(itemPath).isDirectory()) {
+        const instanceYml = loadInstanceConfig(itemPath);
+        if (instanceYml && instanceYml.component) {
```
medikoo, line 455: Actual logic should be: If there's at least one _component_ project, and there are no other `serlveress.yml` projects.

So technically we should stop iteration only when we approach a folder that has `serlveress.yml` that doesn't reflect a component configuration

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@serverless/components",
-  "version": "2.31.10",
+  "version": "2.31.11",
```
medikoo, line 3: We should confirm with @eahefnawy (and maybe @astuyve ) whether we should do a realese (note that drop of `--all` and `sls init` support was not yet released).

It would also make a minor version at least, altough dropping `--all` feels as breaking. It'll probably be nice to release with v3 (we could also officially drop support for Node.js v8 then)

src/cli/utils.js
hkbarton: ok, I think the part I'm missing here is `there are no other serlveress.yml projects.`

medikoo: Yes, we should not assume it's a multi-template project if e.g. one of the subfolders contains a v1 service.

medikoo: Note that logic in master does that, it automatically says _no_ if it approaches a folder with `serverless.yml` that doesn't have a `component` property

package.json
eahefnawy: yeah why do we need to release this first? We haven't even released the code this PR fixes.

hkbarton:
> Looks good. We don't need to release that now though because the --all flag is not yet published, so it shouldn't affect anyone.
> 
> We could merge it and release it to canary though.

sounds good!

PR #698, opened 2020-07-02
Ux arch improvements

ac360:
## What has been implemented?

This is a large PR that generally tries to make the architecture, particularly error handling consistent, as well as address many user experience issues.

- [x] Adjusted global error handling.  Now shows no stack traces by default.  Shows stack traces when the `SLS_DEBUG` env is set or when using `--debug`.

- [x] Added support for looking up 2 parent directories for .env files, instead of just 1.

- [x] Don't require Tencent/China utils libraries to detect if user is based in china.  Use our own simple utility.

- [x] Updated CLI utilities to be more clear.   It seems like many contributors did not understand these.  Added comments to help with understanding.  **This included changing method names throughout the entire codebase.  Please test to ensure this didn't introduce bugs or regressions**.

- [x] Removed the "create" command, since it doesn't do anything and we now have `init`.

- [x] Fixed Dev mode error handling of failed deployments.  Failed deployments can happen in multiple ways, all of which didn't have error handling and would break the CLI.  Now, when deployments fail in various ways, errors are handled and presented gracefully.

- [x] Fixed many vocabulary and grammatical issues.  Increased use of the word "app" as opposed to "instance".

- [x] Improved the "serverless deploy" command when deploying multiple apps and having some fail.  Now, failed deployments are shown clearly alongside successful deployments.  You also get a useful ending status showing success and failed counts.

- [x] Fixed bug where users would not get the login error/advertisement if they tried to do actions without being logged in.

- [x] Added Dashboard URLs where relevant.


eahefnawy:
@hkbarton we need to test this more before releasing. I'm testing now...

src/cli/commands/registry.js
```diff
@@ -216,17 +222,21 @@ const listFeatured = async (config, cli) => {
...
-  cli.log('Featured Components:');
-  cli.log();
-  cli.log('  express - https://github.com/serverless-components/express');
-  cli.log('  website - https://github.com/serverless-components/website');
-  cli.log('  aws-lambda - https://github.com/serverless-components/aws-lambda');
-  cli.log('  aws-dynamodb - https://github.com/serverless-components/aws-dynamodb');
-  cli.log('  aws-iam-role - https://github.com/serverless-components/aws-iam-role');
-  cli.log('  aws-lambda-layer - https://github.com/serverless-components/aws-lambda-layer');
+  cli.log('Featured:');
+  // cli.log();
```
eahefnawy, line 226: This probably needs to be removed?!

src/cli/commands/registry.js
```diff
@@ -41,9 +49,8 @@ const publish = async (config, cli) => {
 
     // If no serverless.yml and no serverless.component.yml, there is nothing to publish in this cwd
     if (!serverlessFile && !serverlessComponentFile) {
-      return cli.error(
-        "Publish failed. The current working directory does not contain a 'serverless.yml' or 'serverless.component.yml'",
-        true
+      throw new Error(
+        'Publish failed. The current working directory does not contain a "serverless.yml" or "serverless.component.yml"'
```
eahefnawy, line 53: We should no longer talk about `serverless.component.yml`. We've decided to go all-in with `serverless.yml`. We just support it for backward compatibility.

Once this release is announced, we should update the docs to remove everything about `serverless.component.yml`

src/cli/commands/registry.js
```diff
@@ -216,17 +222,21 @@ const listFeatured = async (config, cli) => {
...
+  cli.log('• website - https://github.com/serverless-components/website');
+  cli.log('• graphql - https://github.com/serverless-components/graphql');
+  cli.log('• aws-lambda - https://github.com/serverless-components/aws-lambda');
+  cli.log('• aws-dynamodb - https://github.com/serverless-components/aws-dynamodb');
+  cli.log('• aws-iam-role - https://github.com/serverless-components/aws-iam-role');
+  cli.log('• aws-lambda-layer - https://github.com/serverless-components/aws-lambda-layer');
   cli.log();
   cli.log('Find more here: https://github.com/serverless-components');
   cli.log();
+  cli.log('Run "serverless init <package>" to install');
```
eahefnawy, line 238: These are all component packages. `sls init` would just spit a basic yaml file. Don't we wanna put the templates here instead?

src/cli/commands/runAll.js
```diff
@@ -123,6 +131,29 @@ module.exports = async (config, cli, command) => {
...
+        if (c.outputs) {
+          succeeded.push(c.name);
+        }
+      }
+    }
+
+    if (failed.length) {
+      cli.sessionStop(
+        'error',
+        `Errors: "${command}" ran for ${succeeded.length} apps successfully. ${failed.length} failed.`
```
eahefnawy, line 152: That's a cool experience 👍 ... I find the term "apps" confusing in that context though. All nested template instances are always deployed to a single `app`, otherwise we throw an error as discussed. Why not `"deploy" ran for 2 instances successfully`?

hkbarton:
> @hkbarton we need to test this more before releasing. I'm testing now...

agree, I did some test for CN.

src/cli/commands/registry.js
astuyve: +1

src/cli/commands-cn/dev.js
```diff
@@ -120,24 +120,25 @@ module.exports = async (config, cli) => {
...
       await watcher.close();
     }
-    cli.status('Disabling Dev Mode & Closing', null, 'green');
+    cli.sessionStatus('Disabling Dev Mode & Closing', null, 'green');
     const deployedInstance = await deploy(sdk, instanceYaml, instanceCredentials);
     if (await updateDeploymentStatus(cli, deployedInstance, false)) {
-      cli.close('success', 'Dev Mode Closed');
-    } else {
-      cli.close('error', 'Last deployment fail. Run "serverless deploy" to deploy again.');
+      cli.sessionStop('success', 'Dev Mode Closed');
```
astuyve, line 133: Since `sessionStop` is a void method, can't we just `return cli.sessionStop()` here instead of `return null;`?

src/cli/commands/registry.js
ac360: +1

src/cli/commands/registry.js
ac360: Ok.  Didn't know we got rid of that config file.

src/cli/commands/registry.js
ac360: Let's revisit this shortly.

src/cli/commands/runAll.js
ac360: We're moving away from instances and doubling down on apps.

src/cli/commands-cn/dev.js
ac360: Yeah, let's follow up later on that and discuss that lint rule in the first place :P

astuyve: haha, sounds good!

PR #705, opened 2020-07-08
feat: support auto install npm deps for cn init

hkbarton:
## What has been implemented?

support automatically install npm dependencies for CN init command.

### DIFF
--- a/src/cli/commands-cn/init.js
+++ b/src/cli/commands-cn/init.js
@@ -20,8 +20,8 @@ async function unpack(cli, dir, isTopLevel = false) {
   // Check if the directory contains a serverless.yml/yaml/json/js.
   // If it does, we need to unpack it
   if (getServerlessFilePath(dir) || isTopLevel) {
-    cli.sessionStatus(`Installing node_modules via npm in ${dir}`);
     if (await fse.exists(path.resolve(dir, 'package.json'))) {
+      cli.sessionStatus(`Installing node_modules via npm in ${dir}`);
       try {
         await spawn('npm', ['install'], { cwd: dir });
       } catch (error) {
@@ -75,11 +75,8 @@ module.exports = async (config, cli) => {
     throw new Error(`Serverless Registry Package "${packageName}" does not exist.`);
   }
 
-  let targetName = registryPackage.name;
-  if (config.params && config.params.length > 1) {
-    targetName = config.params[1];
-  }
-  const targetPath = path.resolve(process.cwd(), targetName);
+  const targetName = config.name || registryPackage.name;
+  const targetPath = path.resolve(targetName);
 
   if (registryPackage.type !== 'template') {
     await fse.mkdir(targetPath);
@@ -88,7 +85,7 @@ module.exports = async (config, cli) => {
     await fse.writeFile(envDestination, envConfig);
   } else {
     cli.sessionStatus('Fetching template from registry', packageName);
-    const tmpFilename = path.resolve(process.cwd(), path.basename(registryPackage.downloadKey));
+    const tmpFilename = path.resolve(path.basename(registryPackage.downloadKey));
     await pipeline(got.stream(registryPackage.downloadUrl), fs.createWriteStream(tmpFilename));
 
     cli.sessionStatus('Unpacking your new app', packageName);
--- a/src/cli/commands-cn/init.js
+++ b/src/cli/commands-cn/init.js
@@ -52,37 +52,56 @@ module.exports = async (config, cli) => {
   cli.logLogo();
   cli.log();
 
-  let templateName = config.t || config.template;
-  if (!templateName) {
+  let packageName = config.t || config.template;
+  if (!packageName) {
     if (config.params && config.params.length > 0) {
-      templateName = config.params[0];
+      packageName = config.params[0];
     } else {
-      throw new Error('Need to specify template name. e.g. "serverless init some-template".');
+      throw new Error(
+        'Need to specify component or template name. e.g. "serverless init some-template".'
+      );
     }
   }
 
   const sdk = new ServerlessSDK();
-  const template = await sdk.getPackage(templateName);
-  if (!template || template.type !== 'template') {
-    throw new Error(`Template "${templateName}" does not exist.`);
+  const registryPackage = await sdk.getPackage(packageName);
+  if (!registryPackage) {
+    throw new Error(`Serverless Registry Package "${packageName}" does not exist.`);
+  }
+  // registryPackage.component will be null if component is not found
+  // this is the design for platform API backward compatibility
+  delete registryPackage.component;
+  if (Object.keys(registryPackage).length === 0) {
+    throw new Error(`Serverless Registry Package "${packageName}" does not exist.`);
   }
 
-  const targetPath = path.resolve(process.cwd(), template.name);
+  let targetName = registryPackage.name;
+  if (config.params && config.params.length > 1) {
+    targetName = config.params[1];
+  }
+  const targetPath = path.resolve(process.cwd(), targetName);
 
-  cli.sessionStatus('Fetching template from registry', templateName);
-  const tmpFilename = path.resolve(process.cwd(), path.basename(template.downloadKey));
-  await pipeline(got.stream(template.downloadUrl), fs.createWriteStream(tmpFilename));
+  if (registryPackage.type !== 'template') {
+    await fse.mkdir(targetPath);
+    const envDestination = path.resolve(targetPath, 'serverless.yml');
+    const envConfig = `component: ${packageName}\nname: ${targetName}\ninputs:\n`;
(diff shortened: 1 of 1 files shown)


src/cli/commands-cn/init.js
```diff
@@ -5,14 +5,38 @@
...
 
 const pipeline = promisify(require('stream.pipeline-shim'));
 
+async function unpack(cli, dir, isTopLevel = false) {
+  // Check if the directory contains a serverless.yml/yaml/json/js.
+  // If it does, we need to unpack it
+  if (getServerlessFilePath(dir) || isTopLevel) {
+    cli.sessionStatus(`Installing node_modules via npm in ${dir}`);
+    if (await fse.exists(path.resolve(dir, 'package.json'))) {
+      await spawn('npm', ['install'], { cwd: dir });
```
medikoo, line 25: This will fail if `npm` is not installed. Is this ok?

hkbarton: hmm, good point. I've tried to see what's the error message looks like by changing `npm` to something else(npmd), and I got the message `fullstack › Error: spawn npmd ENOENT`, I guess this is ok? since it tells user `npm` could not start. Also, I think the global version did the same.

medikoo: Yes, I think it'll be nice to wrap it with `try/catch` and in case `error.code == 'ENOENT'` show some user friendly message like: "Ensure to install project dependencies".

Alternatively we can completely ignore it

hkbarton: done

hkbarton:
I've updated this PR to include couple changes for better align with global version of serverless init:

1. support putting package name directly after init, like `sls init fullstack`
2. support init from a component, not just template. if init from a component, like the global version we'll create a default `serverless.yml` file
3. support specify application name through second parameters, like `sls init express my-app`. This will create folder `my-app` instead of `express`, also will put `my-app` as `name` field in `serverless.yml` if init from the component. We don't have the token in the CN version, thus we need this to pass in the application name. In the global version, we use token to store application name in backend and retrieving it back through the init.

cc @medikoo @zongUMR

hkbarton:
> > support specify application name through second parameters, like sls init express my-app
> 
> Usual conventions in CLI programs is to put optionals behind a flag.e.g `sls init express --name my-app`. Such notation will be also more self explanatory.
> 
> Or do we follow already a convention coined in non China platform?

I see, ok, let me update to `--name` then. nonCN version doesn't need this since it's using a token that stores app name associate with it.

src/cli/commands-cn/init.js
```diff
@@ -21,32 +52,53 @@ module.exports = async (config, cli) => {
...
-  zip.extractAllTo(path.resolve(process.cwd(), template.name));
-  await fs.promises.unlink(tmpFilename);
+  if (registryPackage.type !== 'template') {
+    await fse.mkdir(targetPath);
+    const envDestination = path.resolve(targetPath, 'serverless.yml');
+    const envConfig = `component: ${packageName}\nname: ${targetName}\ninputs:\n`;
+    await fse.writeFile(envDestination, envConfig);
+  } else {
+    cli.sessionStatus('Fetching template from registry', packageName);
+    const tmpFilename = path.resolve(process.cwd(), path.basename(registryPackage.downloadKey));
```
medikoo, line 88: Same here

src/cli/commands-cn/init.js
```diff
@@ -5,14 +5,45 @@
...
+const { getServerlessFilePath } = require('../serverlessFile');
+const spawn = require('child-process-ext/spawn');
 
 const pipeline = promisify(require('stream.pipeline-shim'));
 
+async function unpack(cli, dir, isTopLevel = false) {
+  // Check if the directory contains a serverless.yml/yaml/json/js.
+  // If it does, we need to unpack it
+  if (getServerlessFilePath(dir) || isTopLevel) {
+    cli.sessionStatus(`Installing node_modules via npm in ${dir}`);
```
medikoo, line 23: This log should be in _if_ clause

src/cli/commands-cn/init.js
```diff
@@ -5,14 +5,45 @@
...
 const { ServerlessSDK } = require('@serverless/platform-client-china');
+const { getServerlessFilePath } = require('../serverlessFile');
+const spawn = require('child-process-ext/spawn');
 
 const pipeline = promisify(require('stream.pipeline-shim'));
 
+async function unpack(cli, dir, isTopLevel = false) {
+  // Check if the directory contains a serverless.yml/yaml/json/js.
+  // If it does, we need to unpack it
+  if (getServerlessFilePath(dir) || isTopLevel) {
```
medikoo, line 22: Why simply not just check for `package.json` only?

src/cli/commands-cn/init.js
```diff
@@ -5,14 +5,45 @@
...
+    cli.sessionStatus(`Installing node_modules via npm in ${dir}`);
+    if (await fse.exists(path.resolve(dir, 'package.json'))) {
+      try {
+        await spawn('npm', ['install'], { cwd: dir });
+      } catch (error) {
+        cli.logError(
+          'Failed install dependencies, make sure install dependencies manually before deploy.'
+        );
+        return null;
```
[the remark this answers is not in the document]
hkbarton: I think so, it just skips to install node module dependencies for the rest of the subproject. user can still install them manually

src/cli/commands-cn/init.js
```diff
@@ -5,14 +5,45 @@
...
+      }
+    }
+
+    const files = await fse.readdir(dir);
+    for (const file of files) {
```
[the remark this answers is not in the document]
hkbarton: doing parallel will cause status reporting of which project's dependencies are being installed not accurate. e.g. if there are 2 subproject API and frontend, the status will show randomly between API and frontend, it may tell user it's installing deps for API, but actually it's installing deps for frontend

src/cli/commands-cn/init.js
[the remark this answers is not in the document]
medikoo: Indeed, makes sense

src/cli/commands-cn/init.js
```diff
@@ -21,32 +52,53 @@ module.exports = async (config, cli) => {
...
+  if (Object.keys(registryPackage).length === 0) {
+    throw new Error(`Serverless Registry Package "${packageName}" does not exist.`);
   }
 
-  cli.sessionStatus('Downloading', templateName);
-
-  const tmpFilename = path.resolve(process.cwd(), path.basename(template.downloadKey));
-  await pipeline(got.stream(template.downloadUrl), fs.createWriteStream(tmpFilename));
+  const targetName = config.name || registryPackage.name;
+  const targetPath = path.resolve(process.cwd(), targetName);
```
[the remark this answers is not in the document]
hkbarton: done

src/cli/commands-cn/init.js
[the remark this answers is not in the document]
medikoo: Yes, and I think it's not safe for npm installs to go in parallel

src/cli/commands-cn/init.js
hkbarton: done

src/cli/commands-cn/init.js
hkbarton: done

src/cli/commands-cn/init.js
hkbarton: hmm, this logic is copied from the global version, I guess we want to make sure we only install dependencies for the deployable serverless projects(which includes `serverless.yml`)?

medikoo: > hmm, this logic is copied from the global version

I'm not sure if there's any big thought behind it over there.

I personally can't think of an use case of having multiple packages in context of one component instances so it's probably not a use case to consider. Still I also don't see any harm in issuing `npm install` in every folder that has `package.json` (without double confirming on existence of `serverless.yml` in same folder), and that will simplify this logic significantly (make code lighter and more maintainable)

Anyway, still if we want to stick to confirming both, then this recursive function could be designed better (currently it shares anti-pattern: https://ariya.io/2011/08/hall-of-api-shame-boolean-trap, and more optimal would be to have `isTopLevel` first in if clause).

What about something as:

```javascript
async function unpack(cli, dir, { isComponentRoot }) {
  if (isComponentRoot) {
    // Eventually npm install
  }
  for (...) {
    cost subDirPath = path.resolve(dir, file)
    await unpack(cli, subDirPath, { isComponentRoot: getServerlessFilePath(subDirPath) });
  }
}
```

hkbarton: agree, I've simplified this part of the logic

hkbarton:
@medikoo @zongUMR updated PR with the simplification of installing deps.

PR #707, opened 2020-07-11
replace src with main in docs

danielcondemarin:
## What has been implemented?

A fix for the docs that can mislead developers building components 

## Steps to verify

N/A

## Todos:

- [ ] Write tests
- [ ] Run Prettier

### DIFF
--- a/README.md
+++ b/README.md
@@ -636,7 +636,7 @@ keywords: aws, serverless, express # Optional. The keywords of the Component to 
 repo: https://github.com/owner/project # Optional. The code repository of the Component
 license: MIT # Optional. The license of the Component code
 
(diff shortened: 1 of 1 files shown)


eahefnawy:
The `src` property now replaces `main` in the latest version of the CLI. We are trying to consolidate naming patterns across component and instance configuration. The main property should still work for backward compatibly, but I suggest you use src moving forward.

Did you have any issues when using `src`? And what version of the CLI were you running? 🤔

danielcondemarin:
> Did you have any issues when using src? And what version of the CLI were you running? 🤔

I'll get back to you on that later. Can't remember exactly which error was.

PR #708, opened 2020-07-13
Reorder properties on init

eahefnawy:
- Makes sure org and app props are at the top
- Removes yaml dump, and instead saves files content as a string to preserve comments.
- Fixes the `inputs: null` issue caused by yaml dump
- Removes excessive newlines

### DIFF
--- a/src/cli/serverlessFile.js
+++ b/src/cli/serverlessFile.js
@@ -1,9 +1,8 @@
 'use strict';
 
 const path = require('path');
-const yaml = require('js-yaml');
 
-const { writeFile } = require('fs-extra');
+const { writeFile, readFile } = require('fs-extra');
 const { fileExistsSync, readAndParseSync } = require('./utils');
 /**
  *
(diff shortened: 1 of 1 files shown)


src/cli/serverlessFile.js
```diff
@@ -100,15 +99,26 @@ const writeServerlessFile = async (cli, servicePath, ymlObject) => {
...
+      // we want the org at the top, so maintaining this order is important
+      if (serviceName) {
+        slsConfig = `service: ${serviceName}\n${slsConfig}`;
+      }
+      if (appName) {
+        slsConfig = `app: ${appName}\n${slsConfig}`;
+      }
+      if (orgName) {
+        slsConfig = `org: ${orgName}\n${slsConfig}`;
+      }
```
medikoo, line 116: Is it guaranteed that `slsConfig` doesn't have this properties set initially?

eahefnawy: yes, they are automatically removed on publishing

PR #709, opened 2020-07-14
Adds a simple help screen

eahefnawy:
If you run `serverless help` or `serverless --help` or `serverless -h` in a components context, you'll see the following:

Screen Shot 2020-07-14 at 4 41 36 PM

### DIFF
--- a/src/legacy.js
+++ b/src/legacy.js
@@ -19,19 +19,6 @@ const componentKeywords = new Set(['registry', 'init', 'publish']);
 const runningComponents = () => {
   const args = minimist(process.argv.slice(2));
 
-  const isRunningHelpOrVersion =
-    process.argv[2] === 'version' ||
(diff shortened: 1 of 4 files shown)


src/legacy.js
```diff
@@ -19,19 +19,6 @@ const componentKeywords = new Set(['registry', 'init', 'publish']);
 const runningComponents = () => {
   const args = minimist(process.argv.slice(2));
 
-  const isRunningHelpOrVersion =
```
medikoo, line 22: Let's maybe also support `--help-components` which will enforce help for components in non component contexts.

I will then also document existence of it in V1 help, so it'll be easy to redirect users which are looking for help specifically on components

eahefnawy: Sounds good 👌

eahefnawy: Done

PR #710, opened 2020-07-15
Release 2.32.0

hkbarton:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@serverless/components",
-  "version": "2.31.12",
+  "version": "2.31.13",
```
medikoo, line 3: Let's make it `2.32.0`, as it's not just bug fixes that it covers

PR #711, opened 2020-07-15
Make sure v1 variables are not used

eahefnawy:


src/cli/utils.js
```diff
@@ -88,6 +88,17 @@ const readAndParseSync = (filePath, options = {}) => {
   return contents.toString().trim();
 };
 
+const validateAgainstV1Variables = (variable) => {
+  const v1Variables = ['self', 'opt', 'sls', 'cf', 's3', 'ssm', 'file'];
```
eahefnawy, line 92: @medikoo anything else I might be missing?

medikoo: Here you have all listed: https://github.com/serverless/serverless/blob/f7dd0334e289f3c3622d0a156ceac588593948cf/lib/classes/Variables.js#L52-L67

There are also `param` and `output` in dashboard plugin, but I take we support them here.

eahefnawy: don't we have something called `${secret:}` too in dashboard plugin? 🤔

eahefnawy: yeah seem to match the list I followed here:
https://www.serverless.com/framework/docs/providers/aws/guide/variables/

src/cli/utils.js
```diff
@@ -88,6 +88,17 @@ const readAndParseSync = (filePath, options = {}) => {
...
+
+  for (const v1Variable of v1Variables) {
+    const v1VariableRegex = new RegExp(`\\\${${v1Variable}:(\\w*[\\w.-_]+)}`, 'g');
```
medikoo, line 95: I think it's equivalent to: `\\\${${v1Variable}:([\\w.-_]+)}` (?)

eahefnawy: Oh yeah that works 👌

PR #713, opened 2020-07-17
types proposal

ac360:


README.md
```diff
@@ -743,156 +707,222 @@ class MyComponent extends Component {
...
+name: express
+version: 1.5.7
+org: serverlessinc
+description: Deploy a serverless Express.js application onto AWS Lambda and AWS HTTP API.
+
+actions:
+  # deploy action
+  deploy:
+    # deploy action definition
+    definition: Deploy your Express.js application to AWS Lambda, AWS HTTP API and more.
```
eahefnawy, line 737: as mentioned on Slack, I'd change that to `description`

README.md
```diff
@@ -743,156 +707,222 @@ class MyComponent extends Component {
...
+      #
+      #
+      # Primitive Types
+      # These cover standard data types, like "string", "number", "object", etc.
+      #
+      #
+
+      # Type: string
+
+      - name: name # The name of the parameter
```
eahefnawy, line 768: I'd also go with an object instead of an array with the value of the `name` property as the key for each input type.

README.md
```diff
@@ -743,156 +707,222 @@ class MyComponent extends Component {
...
+      # Type: array
+
+      - name: mappingTemplates # The name of the parameter
+        type: array # The type
+        # Optional
+        required: true # Defaults to required: false
+        description: The mapping templates for your GraphQL endpoints. # A description of this input
+        min: 1 # Minimum array items
+        max: 10 # Max array items
+        items:
```
eahefnawy, line 825: I'd rename `items` to `allow`. This way it's clear that it's an array of allowed types for any of the array items.

README.md
```diff
@@ -743,156 +707,222 @@ class MyComponent extends Component {
...
+        dist: # The directory containing your built source code which you wish to upload.
+        exclude: # An array of glob patterns of files/paths to exclude
+          - .env # exclude .env file in ./src
+          - '.git/**' # exclude .git folder and all subfolders and files inside it
+          - '**/*.log' # exclude all files with .log extension in any folder under the ./src
+
+      # Type: env
+      # This Type is for an object of key-value pairs meant to contain sensitive information.  By using it, the Serverless Framework will treat this data more securely.
+
+      - name: env # The name "env" is reserved for this Special Type.  Your params can only have one of these.
```
eahefnawy, line 869: why are we restricting to only one env type for inputs? I can't think of a use case where we might need more, but I don't see the point of restriction it either 🤔 ... Maybe a different input could use it for other types of secrets.

PR #720, opened 2020-07-21
Validate node_modules exists when publishing components

eahefnawy:
Otherwise, the Lambda function would return an `internal server error` and it'd be hard to figure out what happened.

### DIFF
--- a/src/cli/utils.js
+++ b/src/cli/utils.js
@@ -674,6 +674,32 @@ const isChinaUser = () => {
   return result;
 };
 
+/**
+ * Makes sure user ran "npm install" if package.json with dependencies was found
+ */
+const validateNodeModules = async (directory) => {
+  const srcPath = path.resolve(process.cwd(), directory);
+  const packageJsonPath = path.resolve(srcPath, 'package.json');
+
+  // only validate node modules if package.json exists
+  if (await fileExists(packageJsonPath)) {
+    const packageJson = readAndParseSync(packageJsonPath);
+
+    const hasDependencies = Object.keys(packageJson.dependencies).length > 0;
+
+    // only validate node modules if there are dependencies in package.json
+    if (hasDependencies) {
+      const nodeModulesExists = await fse.pathExists(path.resolve(srcPath, 'node_modules'));
+
+      if (!nodeModulesExists) {
+        throw new Error(
+          'node_modules was not found in the current working directory, or the "src" directory you specified. Did you run "npm install"?'
+        );
+      }
+    }
+  }
+};
+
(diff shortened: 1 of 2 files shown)


eahefnawy:
> Do we understand what actually happens and why internal server error is returned?

Yeah because the user code is trying to load modules that don't exist, which is why we should validate it before publishing.

eahefnawy:
Ok, I forgot to clarify that this only happens if you forgot to run `npm install` prior to publishing if you're using any packages. So it's expected and not a bug. Here we just give early feedback.

medikoo:
> Yeah because the user code is trying to load modules that don't exist

Load in what way? Via `require` ?  Why it's an `internal server error` and not a meaningful error that's thrown?

eahefnawy:
Yeah, we probably don't try/catch that deployment error. Nonetheless, it should be validated on the CLI during publishing, not during deploying, for the earliest feedback, hence this change.

medikoo:
I'm still wondering why we expose it as "internal server error" (?) Invalid requires are reported  in lambdas  with "module initialization error: Error".

Doesn't it signal a bug in error handling on our side?

I understand that this patch will fix this single error case, still I think it'll be good to investigate it further, to fully understand why actually it is reported with meaningless "internal server error" message and why we do not have access to stack trace that'll give us valuable debugging information (?)

eahefnawy:
> I'm still wondering why we expose it as "internal server error" (?) Invalid requires are reported in lambdas with "module initialization error: Error".

We don't expose an internal server error. It's probably APIG that does that because it's unhandled. We could handle it, but that's beyond the scope of this PR. I'm already tracking that separate issue here (which applies to other errors too):
https://github.com/serverlessinc/platform/issues/51

Meanwhile, let's merge this in until we figure this out. It provides better dx/feedback for this specific case regardless of how we end up handling different errors.

medikoo:
> It's probably APIG that does that

ok, I missed that part.

> I'm already tracking that separate issue here (which applies to other errors too):
serverlessinc/platform#51

👍

src/cli/utils.js
```diff
@@ -674,6 +674,32 @@ const isChinaUser = () => {
   return result;
 };
 
+/**
+ * Makes sure user ran "npm install" if package.json with dependencies was found
+ */
+const validateNodeModules = async (directory) => {
+  const srcPath = path.resolve(process.cwd(), directory);
```
medikoo, line 681: nit: with `path.resolve` paths are resolved naturally against `process.cwd()`, so that first argument can be skipped

PR #726, opened 2020-07-24
feat: support debug mode for CN platform

hkbarton:
## What has been implemented?

support `--debug` for CN user.

### DIFF
--- a/src/cli/commands-cn/run.js
+++ b/src/cli/commands-cn/run.js
@@ -6,7 +6,7 @@
 
 const path = require('path');
 const { runningTemplate } = require('../utils');
-const { ServerlessSDK } = require('@serverless/platform-client-china');
+const { ServerlessSDK, utils: tencentUtils } = require('@serverless/platform-client-china');
 const utils = require('./utils');
 const runAll = require('./runAll');
 const chalk = require('chalk');
@@ -46,8 +46,15 @@ module.exports = async (config, cli, command) => {
   const instanceCredentials = await utils.loadInstanceCredentials(instanceYaml.stage);
 
   // initialize SDK
+  const orgUid = await tencentUtils.getOrgId();
   const sdk = new ServerlessSDK({
+    accessKey: tencentUtils.buildTempAccessKeyForTencent({
+      SecretId: process.env.TENCENT_SECRET_ID,
+      SecretKey: process.env.TENCENT_SECRET_KEY,
+      Token: process.env.TENCENT_TOKEN,
+    }),
     context: {
+      orgUid,
       orgName: instanceYaml.org,
     },
   });
@@ -59,6 +66,18 @@ module.exports = async (config, cli, command) => {
   const options = {};
   options.debug = config.debug;
   options.dev = config.dev;
+
+  // Connect to Serverless Platform Events, if in debug mode
+  if (options.debug) {
+    await sdk.connect({
+      filter: {
+        stageName: instanceYaml.stage,
+        appName: instanceYaml.app,
+        instanceName: instanceYaml.name,
+      },
+      onEvent: utils.handleDebugLogMessage(cli),
+    });
+  }
 
   let deferredNotificationsData;
   if (command === 'deploy') {
@@ -116,5 +135,7 @@ module.exports = async (config, cli, command) => {
   cli.sessionStop('success', 'Success');
 
   if (deferredNotificationsData) printNotification(cli, await deferredNotificationsData);
+
+  sdk.disconnect();
   return null;
 };
--- a/src/cli/commands-cn/runAll.js
+++ b/src/cli/commands-cn/runAll.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const { ServerlessSDK } = require('@serverless/platform-client-china');
+const { ServerlessSDK, utils: tencentUtils } = require('@serverless/platform-client-china');
 const {
   getOutputs,
   getAllComponents,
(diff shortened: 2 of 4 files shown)


src/cli/commands-cn/run.js
```diff
@@ -60,6 +67,37 @@ module.exports = async (config, cli, command) => {
...
+      },
+      onEvent: (evt) => {
+        if (evt.event !== 'instance.run.logs') {
+          return;
+        }
+        if (evt.data.logs && Array.isArray(evt.data.logs)) {
+          evt.data.logs.forEach((log) => {
+            // Remove strange formatting that comes from stderr
+            if (typeof log.data === 'string' && log.data.startsWith("'")) {
+              log.data = log.data.substr(1);
```
medikoo, line 86: `substr` is deprecated string method. Best is to just rely on `slice` (which shares same rules as `arr.slice` hence easy to remember)

src/cli/commands-cn/run.js
```diff
@@ -60,6 +67,37 @@ module.exports = async (config, cli, command) => {
...
+        instanceName: instanceYaml.name,
+      },
+      onEvent: (evt) => {
+        if (evt.event !== 'instance.run.logs') {
+          return;
+        }
+        if (evt.data.logs && Array.isArray(evt.data.logs)) {
+          evt.data.logs.forEach((log) => {
+            // Remove strange formatting that comes from stderr
+            if (typeof log.data === 'string' && log.data.startsWith("'")) {
```
medikoo, line 85: Can `log.data` be not of string type?

src/cli/commands-cn/run.js
```diff
@@ -60,6 +67,37 @@ module.exports = async (config, cli, command) => {
...
+      filter: {
+        stageName: instanceYaml.stage,
+        appName: instanceYaml.app,
+        instanceName: instanceYaml.name,
+      },
+      onEvent: (evt) => {
+        if (evt.event !== 'instance.run.logs') {
+          return;
+        }
+        if (evt.data.logs && Array.isArray(evt.data.logs)) {
```
medikoo, line 82: Could be just `if (Array.isArray(evt.data.logs)) {`

src/cli/commands-cn/runAll.js
```diff
@@ -46,10 +53,38 @@ module.exports = async (config, cli, command) => {
...
+            }
+            if (typeof log.data === 'string' && log.data.endsWith('\\n')) {
+              log.data = log.data.substring(0, log.data.length - 2);
+            }
+            cli.log(log.data);
+          });
+        }
+      },
+    });
+  }
```
medikoo, line 86: This looks as copy/pasted logic from `run`. Can we put it in utils, and reuse?

src/cli/commands-cn/run.js
hkbarton: done

src/cli/commands-cn/run.js
hkbarton: seems like it should be always string

src/cli/commands-cn/run.js
hkbarton: done

src/cli/commands-cn/runAll.js
hkbarton: done

src/cli/commands-cn/run.js
```diff
@@ -116,5 +135,7 @@ module.exports = async (config, cli, command) => {
   cli.sessionStop('success', 'Success');
 
   if (deferredNotificationsData) printNotification(cli, await deferredNotificationsData);
+
+  sdk.disconnect();
```
medikoo, line 139: What if connection was not initialized (with `options.debug` being falsy), wouldn't it crash?

src/cli/commands-cn/runAll.js
```diff
@@ -95,4 +111,7 @@ module.exports = async (config, cli, command) => {
   cli.sessionStop('success', 'Success');
 
   if (deferredNotificationsData) printNotification(cli, await deferredNotificationsData);
+
+  sdk.disconnect();
```
medikoo, line 115: Same here

src/cli/commands-cn/utils.js
```diff
@@ -227,6 +227,29 @@ const setInputsForCommand = (instanceYaml, command, config) => {
...
+      evt.data.logs.forEach((log) => {
+        // Remove strange formatting that comes from stderr
+        if (log.data.startsWith("'")) {
+          log.data = log.data.slice(1);
+        }
+        if (log.data.endsWith("'")) {
+          log.data = log.data.substring(0, log.data.length - 1);
+        }
+        if (log.data.endsWith('\\n')) {
+          log.data = log.data.substring(0, log.data.length - 2);
```
medikoo, line 245: nit: for consistency it'll be nice to rely on `slice` here as well (it'll even be simpler)

src/cli/commands-cn/run.js
hkbarton: it won't, disconnect is safe: https://github.com/serverlessinc/platform-cn/blob/master/sdk/src/Connection.js#L81

src/cli/commands-cn/utils.js
hkbarton: done

src/cli/commands-cn/utils.js
```diff
@@ -227,6 +227,29 @@ const setInputsForCommand = (instanceYaml, command, config) => {
...
+      return;
+    }
+    if (Array.isArray(evt.data.logs)) {
+      evt.data.logs.forEach((log) => {
+        // Remove strange formatting that comes from stderr
+        if (log.data.startsWith("'")) {
+          log.data = log.data.slice(1);
+        }
+        if (log.data.endsWith("'")) {
+          log.data = log.data.slice(0, log.data.length - 1);
```
medikoo, line 242: nit: It can be just `log.data.slice(0, -1);`

hkbarton: done

PR #730, opened 2020-07-29
Introduce better type system errors and other minor fixes

ac360:
This introduces error handling for types, as well as fixes some `undefined` statements that occurred as a result of us changing our event data format.

* The introduced error handling method will not work however, due to its inability to work with China-based users.  It needs to be reworked before this can be pushed.

Screen Shot 2020-07-28 at 7 56 51 PM copy


src/cli/CLI.js
```diff
@@ -311,7 +328,7 @@ class CLI {
   logLogo() {
     let logo = os.EOL;
     logo += 'serverless';
-    logo += red(' ⚡');
+    logo += red(' ⚡ ');
     logo += 'framework';
```
eahefnawy, line 332: This renders differently on different CLIs 😞

src/cli/CLI.js
```diff
@@ -106,30 +106,29 @@ class CLI {
...
       color = green;
     }
-
-    // Clear any existing content
-    process.stdout.write(ansiEscapes.cursorLeft);
-    process.stdout.write(ansiEscapes.eraseDown);
-
-    // Render stack trace (if debug is on)
-    if (reason === 'error') {
-      this.logErrorStackTrace(message);
```
eahefnawy, line 128: we will no longer render the stack trace in debug mode? that was /super helpful for component development/authors.

ac360: This is still included, but elsewhere.

src/cli/CLI.js
ac360: Yep.  there is no correct way spacing, unfortunately.

ac360:
This has been refactored to not use a custom error object.

Instead, the main point of customization is the `logError()` method in the `CLI.js` file, and everything should use `logError()`. 
 That is where we can differentiate the experience between SFGlobal and SFChina.  Simply pass in an error with the supported, extra properties to have those be displayed.  Defaults should also be set in the `logError()` utility.

Keep in mind, the SFGlobal has been updated slightly to return more data from in API Error Responses to pass through to the CLI to help users.  The standards for designing this are here:   https://github.com/serverlessinc/company/tree/master/docs/standards#api-errors

This code is somewhat hacky, but it should suffice for now.

- [ ] For SFChina, defaults must be added to `logError()`
- [ ] For SFChina, the `deploy` `deployAll` `dev` commands should add custom error info where helpful
- [ ] For SFChina, the `executeGraph` util method must also be customized, if necessary

@hkbarton What do you think of this and can we merge this as-is asap, without affecting the SFChina experience?

src/cli/CLI.js
```diff
@@ -180,28 +180,81 @@ class CLI {
...
+          : `  Documentation: https://github.com/serverless/components ${os.EOL}`;
+      }
+      if (error.support !== false) {
+        error.support = `  Support: https://app.serverless.com/support ${os.EOL}`;
+      }
+      if (error.chat !== false) {
+        error.chat = `  Slack: https://www.serverless.com/slack/ ${os.EOL}`;
+      }
+    } else {
+      if (error.documentation !== false) {
```
hkbarton, line 213: why do we check `!== false` instead of check `if (error.documentation)` ? does this mean we have to pass in `false` specifically to hide doc URLs, and it will render `undefined` if we don't add anything

src/cli/commands/dev.js
```diff
@@ -12,7 +12,30 @@ const {
   isLoggedInOrHasAccessKey,
 } = require('./utils');
 
+/**
+ * Deploy helper function
+ */
+const deploy = async (sdk, cli, instanceYaml, instanceCredentials) => {
+  let result;
+  try {
+    result = await sdk.deploy(instanceYaml, instanceCredentials);
```
astuyve, line 21: Dropped options here, we lost devMode

src/cli/commands/dev.js
```diff
@@ -225,13 +239,8 @@ module.exports = async (config, cli) => {
 
   cli.watcher.on('ready', async () => {
     cli.sessionStatus('Initializing Dev Mode', null, 'green');
-
-    try {
-      await sdk.deploy(instanceYaml, instanceCredentials, { dev: true });
```
astuyve, line 230: Dropped dev

PR #735, opened 2020-07-31
feat: Add templates to CLI onboarding

### LINKED ISSUE #481: Stage Support
What are the best practices to create multiple environments using serverless components? For example, I wanted to configure a website for my prod, staging, dev environments. Not sure whether we can use stage option likes `serverless deploy --stage prod` command

zongUMR:
## What has been implemented?

1. Get data from registry api
2. Users can search for templates by entering name of template
3. Same behavior with `sls init` after choosing a template

Closes https://github.com/serverless/roadmap-tencent/issues/481

## Steps to verify

- Run the local `bin` binary in a non-serverless project 
- Select templates from list
- Input project name
- User can deploy project immediately after creating it successfully

## Todos:

- [ ] Support inputs
- [ ] Support multi-language runtime for templates, like `SCF`

### DIFF
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -20,6 +20,31 @@ module.exports = async () => {
   const instanceConfig = loadInstanceConfig(process.cwd());
   const stage = args.stage || (instanceConfig && instanceConfig.stage) || 'dev';
 
+  const params = [];
+  if (args._[1]) {
+    params.push(args._[1]);
+  }
+  if (args._[2]) {
+    params.push(args._[2]);
+  }
+  if (args._[3]) {
+    params.push(args._[3]);
+  }
+  if (args._[4]) {
+    params.push(args._[4]);
+  }
+
+  const config = { ...args, params };
+  if (config._) {
+    delete config._;
+  }
+
+  config.platformStage = process.env.SERVERLESS_PLATFORM_STAGE || 'prod';
+  config.debug = process.env.SLS_DEBUG || !!args.debug;
+
+  // Initialize CLI utilities
+  const cli = new CLI(config);
+
   /**
    * Load environment variables from .env files, 2 directories up
    * Nearest to current working directory is preferred
@@ -70,7 +95,7 @@ module.exports = async () => {
    */
   if (process.argv.length === 2 && isChinaUser() && !(await isProjectPath(process.cwd()))) {
     // Interactive onboarding
-    return require('./interactive-onboarding/cn')();
+    return require('./interactive-onboarding/cn')(config, cli);
   }
 
   let commands;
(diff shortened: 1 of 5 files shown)


hkbarton:
great work @zongUMR !

src/cli/index.js
```diff
@@ -99,36 +124,11 @@ module.exports = async () => {
...
-  const config = { ...args, params };
-  if (config._) {
-    delete config._;
-  }
-
-  config.platformStage = process.env.SERVERLESS_PLATFORM_STAGE || 'prod';
-  config.debug = process.env.SLS_DEBUG || !!args.debug;
-
   // Add stage environment variable
   if (args.stage && !process.env.SERVERLESS_STAGE) {
```
hkbarton, line 128: should we move those lines up as well?

zongUMR: We can, but I actually only need the `cli` for the onboarding process, will move them up

src/cli/interactive-onboarding/cn.js
```diff
@@ -1,45 +1,31 @@
...
     await inquirer.prompt({
       // EN: What do you want to make?
       message: '请选择你希望创建的 Serverless 应用',
-      type: 'list',
+      type: 'autocomplete',
       name: 'projectType',
-      choices: initializeProjectChoices,
+      source: (_, input) => {
+        if (input !== undefined) {
+          return Promise.resolve(choices.filter((choice) => choice.name.includes(input)));
```
medikoo, line 25: Let's not use `Promise.resolve`, but simply return results (if for some reason only _promise_ is accepted as function return value, simply mark function as _async_)

src/cli/interactive-onboarding/cn.js
```diff
@@ -1,45 +1,31 @@
...
   (
     await inquirer.prompt({
       // EN: What do you want to make?
       message: '请选择你希望创建的 Serverless 应用',
-      type: 'list',
+      type: 'autocomplete',
       name: 'projectType',
-      choices: initializeProjectChoices,
+      source: (_, input) => {
+        if (input !== undefined) {
```
medikoo, line 24: More natural would be to `input != null` (it discards both `undefined` and `null`)

src/cli/interactive-onboarding/cn.js
zongUMR: https://github.com/mokkabonna/inquirer-autocomplete-prompt#options, `source` must return a `promise` here, can also refer to the example: https://github.com/mokkabonna/inquirer-autocomplete-prompt/blob/master/example.js. I try to return the result directly, it will throw an error, it looks like the `source` will use `.then` method inside.

medikoo: Ok, so let's just use _async_ function (using `Promise.resolve` is an anti-pattern, makes implementation unnecessary noisy)

src/cli/interactive-onboarding/cn.js
zongUMR: Maybe if(!!input) is a better way?

medikoo: If `input` is an optional _object_ , then yes simply `if (input)` will do (no need for `!!`)

PR #739, opened 2020-08-04
Fix regression where sls publish is broken

astuyve:
## What has been implemented?

Fixes a bug introduced in #735 where `sls publish` was broken

## Steps to verify

- Pull this branch, run `sls publish`
- `sls registry` still shows the registry items
- `sls registry publish` still publishes

### DIFF
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -119,7 +119,7 @@ module.exports = async () => {
    */
   if (command === 'publish') {
     command = 'registry';
-    config.params.push('publish');
+    config.params.unshift('publish');
   }
 
   /**
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -119,7 +119,7 @@ module.exports = async () => {
    */
   if (command === 'publish') {
     command = 'registry';
-    args._[1] = 'publish';
+    config.params.push('publish');
   }
 
   /**


src/cli/index.js
```diff
@@ -119,7 +119,7 @@ module.exports = async () => {
    */
   if (command === 'publish') {
     command = 'registry';
-    args._[1] = 'publish';
+    config.params.push('publish');
```
medikoo, line 122: Initial implementation converted `sls publish foo` to  `sls registry publish`, while this update converts `sls publish foo` to `sls registry foo publish`.

Probably most accurate would be, if we'd convert `sls publish foo` to `sls registry publish foo` (?)

astuyve: Ah, good point. I will use unshift instead of push.

astuyve: I don't believe it matters, as we don't allow other params for the registry command. (--dev is a flag, not a param).
But best to keep it as accurate as possible.

medikoo: > I don't believe it matters, as we don't allow other params for the registry command

Yes, by spec we do not support any more arguments, but if someone by mistake runs `sls publish component-name` it'll be nice if `sls publish` is run and not `sls component-name` (which may collide with other command names)

PR #740, opened 2020-08-05
Introduce changelog

medikoo:
- Introduce `CHANGELOG.md` file
- Automate publication of release notes in CI:
  - PR build: if version is bumped it is confirmed that release notes were added to CHANGELOG.md (if not validation will fail)
  - Publication build: After successful publication realase notes are read from CHANGELOG.md and published to GitHub


medikoo:
> so if I update CHANGELOG.md alongside with bumping version in package.json, the release tag that being created automatically will include my CHANGELOG.md update right?

After tag is created, the publication build will jump in, which will publish new version and read
release notes from CHANGELOG.md and publish them to GitHub (to release tag)

medikoo:
This is how it works already in https://github.com/serverless/enterprise-plugin/

PR #754, opened 2020-08-13
Pass the actual directory, instead of process.cwd

astuyve:
## What has been implemented?

Passes the newly created directory instead of relying on process.cwd()

### DIFF
--- a/src/cli/commands/init/unpacker.js
+++ b/src/cli/commands/init/unpacker.js
@@ -3,7 +3,11 @@ const fs = require('fs-extra');
 const spawn = require('child-process-ext/spawn');
 const { writeMainAttrs, getServerlessFilePath } = require('../../serverlessFile');
 const path = require('path');
-const { legacyLoadComponentConfig, legacyLoadInstanceConfig } = require('../../utils');
+const {
+  legacyLoadComponentConfig,
+  legacyLoadInstanceConfig,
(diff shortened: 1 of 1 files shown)


astuyve:
@ac360 I've logged out the config file before we write the attributes, I think that extra newline needs to be removed on publish:
```js
    // in ServerlessFile.js
    let slsConfig = await readFile(slsConfigPath, 'utf8');
    console.log('SLS CONFIG: ', slsConfig);
```
```yaml
SLS CONFIG:  component: aws-iam-role

name: permissions

inputs:
  name: ${name}-${stage}
  region: us-east-1
  service: lambda.amazonaws.com
  policy:
    # AWS Lambda function containing Express Logs and Assume Role access
    - Effect: Allow
      Action:
        - sts:AssumeRole
        - logs:CreateLogGroup
        - logs:CreateLogStream
        - logs:PutLogEvents
      Resource: "*"
    # AWS DynamoDB Table access
    - Effect: Allow
      Action:
        - dynamodb:DescribeTable
        - dynamodb:Query
        - dynamodb:GetItem
        - dynamodb:PutItem
        - dynamodb:UpdateItem
        - dynamodb:DeleteItem
      Resource:
        - ${output:database.arn}
        - ${output:database.arn}/index/*
```

PR #756, opened 2020-08-13
fix multi service provisioning

ac360:
Currently, the `serverless` command still tries to do a deploy w/ component-related projects, and the command experience when using a project with multiple services doesn't can fail when using a different command than deploy and remove.  This PR tries to fix/improve a lot of this.

### DIFF
--- a/src/cli/commands/run.js
+++ b/src/cli/commands/run.js
@@ -25,8 +25,8 @@ module.exports = async (config, cli, command) => {
     throw new Error('Please log in by running "serverless login"');
   }
 
-  // Check to see if the cwd is a template containing multiple templates
-  if (['deploy', 'remove'].includes(command) && runningTemplate(process.cwd())) {
+  // Check to see if the cwd is a template containing multiple services
+  if (runningTemplate(process.cwd())) {
     return runAll(config, cli, command);
   }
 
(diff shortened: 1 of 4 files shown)


src/cli/index.js
```diff
@@ -110,7 +110,7 @@ module.exports = async () => {
     commands = require('./commands');
   }
 
-  let command = args._[0] || 'deploy';
+  let command = args._[0] || null;
```
eahefnawy, line 113: doesn't that kill the single command deployment though? No more `serverless` to deploy? 😢

src/cli/commands/run.js
```diff
@@ -25,8 +25,8 @@ module.exports = async (config, cli, command) => {
     throw new Error('Please log in by running "serverless login"');
   }
 
-  // Check to see if the cwd is a template containing multiple templates
-  if (['deploy', 'remove'].includes(command) && runningTemplate(process.cwd())) {
+  // Check to see if the cwd is a template containing multiple services
+  if (runningTemplate(process.cwd())) {
```
eahefnawy, line 29: As raised by @medikoo  on a previous PR, checking for templates is quite expensive, which is why we are limiting that check to just the `deploy` and `remove` commands. Do we really want to make that check on every framework command?

eahefnawy:
Closing this PR for the sake of #829, which addresses the same issues except for template checks as discussed.

PR #758, opened 2020-08-13
Adds unpublish command

eahefnawy:


src/cli/commands/registry.js
```diff
@@ -149,6 +149,100 @@ const publish = async (config, cli) => {
...
+  }
+
+  // Disable timer
+  config.timer = false;
+
+  // Start CLI persistance status
+  cli.sessionStart('Initializing');
+
+  // Get access key
+  const accessKey = await getAccessKey();
```
astuyve, line 172: These could be combined into one parallel operation with `await Promise.all([])`

eahefnawy:
@astuyve thanks, done!

Issue #506, opened 2019-10-09
Using IAM Role instead of IAM Users


sonneym:
I might be wrong, but when I checked the serverless components are deployed only with user's AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY.  It would be nice to be able to support IAM Role and AWS_SESSION_TOKEN.   IAM roles usage is a nice best practice, instead of using permanent IAM User credentials.

eahefnawy:
That makes sense. We will be updating how we handle credentials soon as we're working on a remote deployment engine. We have support for this planned.

lytvynenko:
Hi @eahefnawy! Any updates related to this issue?

daaru00:
Hi all,

could it make sense to directly use the credentials managed by the [Serverless Dashboard](https://app.serverless.com)? when connecting an environment it reports:
> Setup your AWS account with an IAM role so Serverless can deploy directly to your account for this stage

it works for base version of Serverless but not for components, it would be nice to manage them in the same way without the need to manage credentials locally.

In my use case I just can't use the IAM user credentials with AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY. We use the assume role feature to jump into organization's accounts and developers don't have the ability to create new users inside "child" account. Basically it is impossible for us to deploy on these infrastructures.

PR #768, opened 2020-08-18
Add support for AWS profile with role_arn configuration

daaru00:
## What has been implemented?

**Allow usage of symbolic link for AWS credentials and other files.**

For additional security, using Linux, is a good practice to link files with sensitive content (like credentials) into [Encrypted Private Directory](https://help.ubuntu.com/community/EncryptedPrivateDirectory). These changes allow to use this configurations without fail `fileExistsSync()` check.

**Add [aws-sdk](https://www.npmjs.com/package/aws-sdk) package as dependency**

This is required in order to [assume a role](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html) using AWS STS client.

**Makes `loadInstanceCredentials()` asynchronous**

If I have not misunderstood the `loadInstanceCredentials()` function is always called with await operator even though the function was not declared as async. These change will force this declaration in order to wait AWS STS call to retrieve credentials.

**Add support for AWS Profile with assume role functionality**

Closes issue #506

## Steps to verify

- Add into `~/.aws/credentials` a profile section with role_arn
```
[mycompanyprofile]
aws_access_key_id = xxxxxxxxxxxxxxxxxxxx
aws_secret_access_key = xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx

[myprofile]
role_arn = arn:aws:iam::000000000000:role/MyOrganizationAccountAccessRole
source_profile = mycompanyprofile
```
- Set environment variable `AWS_PROFILE` or add it into `.env` file
```
AWS_PROFILE=myprofile
```
- Run `serverless deploy` command

## Todos:

- [ ] Write tests
- [x] Run Prettier

### DIFF
--- a/src/cli/commands/utils.js
+++ b/src/cli/commands/utils.js
@@ -7,6 +7,7 @@
 const args = require('minimist')(process.argv.slice(2));
 const path = require('path');
 const os = require('os');
+const https = require('https');
 const {
   readConfigFile,
   writeConfigFile,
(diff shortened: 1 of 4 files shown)


daaru00:
Hi @eahefnawy,

> any chance you could get this to work without depending on the aws-sdk?

I removed the `aws-sdk` dependency and used [aws4](https://www.npmjs.com/package/aws4) to sign request with [Signature Version 4](https://docs.aws.amazon.com/general/latest/gr/signature-version-4.html). This kind of signature is too complicated to replicate, and doing the whole process from scratch doesn't seem like a good idea..

Is this solution good enough for you?

src/cli/commands/utils.js
```diff
@@ -124,9 +209,9 @@ const loadAwsCredentials = () => {
  * @param {*} stage
  */
 
-const loadInstanceCredentials = () => {
+const loadInstanceCredentials = async () => {
```
eahefnawy, line 212: I realize why you had to make that method async, but that could have serious implications on the codebase where this method might be called. I need to search and review the codebase to make sure there are no calls to this method that we've missed.

eahefnawy (commented):
@daaru00 thank you for making that change. It must've not been easy 😊 

It looks good, I'm just a bit worried about the async change and I wanna test it a bit.

src/cli/commands/utils.js
daaru00: absolutely understandable, definitely needs some testing :+1: let me know

eahefnawy:
@daaru00 I checked the codebase and tested this with multiple scenarios. Couldn't break it. Good job 👍 

If you could just add docs in the [credentials section](https://github.com/serverless/components#credentials), this PR would be G2M!

daaru00:
@eahefnawy that's great! 

I added the documentation with extra section on credentials for AWS, let me know if it's clear enough.

While I was writing some details about session durationI also added the possibility to modify it using a custom environment variable. This in order to to minimize the effect of expired credentials during "watch" command. I guess does not re-execute this portion of code again, am I wrong?

eahefnawy (approved):
Yeah, I think that's fine.

Thanks again @daaru00 ... G2M 🚀

PR #770, opened 2020-08-19
fix: polyfill for fs.promises

zongUMR:
## What has been implemented?

Closes https://github.com/serverless/components/issues/769

## Steps to verify

- Run X
- ...

## Todos:

- [ ] Write tests
- [ ] Run Prettier

### DIFF
--- a/src/cli/commands-cn/init.js
+++ b/src/cli/commands-cn/init.js
@@ -51,7 +51,13 @@ const initTemplateFromCli = async (targetPath, packageName, registryPackage, cli
   cli.sessionStatus('Unpacking your new app', packageName);
   const zip = new AdmZip(tmpFilename);
   zip.extractAllTo(targetPath);
-  await fs.promises.unlink(tmpFilename);
+  // Need to polyfill this promise feature for support of nodejs>=8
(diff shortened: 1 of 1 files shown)


src/cli/commands-cn/init.js
```diff
@@ -51,7 +51,13 @@ const initTemplateFromCli = async (targetPath, packageName, registryPackage, cli
...
   const zip = new AdmZip(tmpFilename);
   zip.extractAllTo(targetPath);
-  await fs.promises.unlink(tmpFilename);
+  // Need to polyfill this promise feature for support of nodejs>=8
+  if (fs.promises) {
+    await fs.promises.unlink(tmpFilename);
+  } else {
+    const unlink = promisify(fs.unlink);
+    await unlink(tmpFilename);
+  }
```
medikoo, line 60: For ease of maintance would be nicer to just use `promisify(fs.unlink)` (maintaining two different versions can be error-prone)

medikoo (approved):

src/cli/commands-cn/init.js
zongUMR: 💯

PR #786, opened 2020-08-27
Creates YAML file in the root of nested templates on init

eahefnawy:
- Creates yaml file in the root of nested templates on init
- Adds app, org & stage properties to the root yaml
- Removes app, org & stage properties from nested components
- Supports inheriting stage and org properties from root yaml


eahefnawy:
The Node 10 check is stuck. Seems to be passing here:
https://github.com/serverless/components/runs/1036432973?check_suite_focus=true

Merging...

PR #792, opened 2020-09-01
throw error when using plugins with components

### LINKED ISSUE #655: Include notice that plugins do not work w/ components
Users are getting confused between Components and SFV.1 and trying to add plugins to them.  We should detect if `plugins` is included and throw an error message if so.

eahefnawy:
Closes #655

### DIFF
--- a/src/cli/commands/utils.js
+++ b/src/cli/commands/utils.js
@@ -329,6 +329,13 @@ const loadVendorInstanceConfig = async (directoryPath, options = { disableCache:
     }
   }
 
+  // give early feedback to people who are trying to use plugins with components
+  if (instanceFile.plugins) {
+    throw new Error(
+      'Serverless Framework plugins are not support by Serverless Components. Please remove the "plugins" property and try again'
+    );
+  }
+
   return instanceFile;
 };
 
--- a/src/cli/commands/utils.js
+++ b/src/cli/commands/utils.js
@@ -332,7 +332,7 @@ const loadVendorInstanceConfig = async (directoryPath, options = { disableCache:
   // give early feedback to people who are trying to use plugins with components
   if (instanceFile.plugins) {
     throw new Error(
-      'Serverless Framework plugins are not support by Serverless Components. Please remove the "plugins" property and try again'
+      'Serverless Framework plugins are not supported by Serverless Components. Please remove the "plugins" property and try again'
     );
   }
 


src/cli/commands/utils.js
```diff
@@ -329,6 +329,13 @@ const loadVendorInstanceConfig = async (directoryPath, options = { disableCache:
     }
   }
 
+  // give early feedback to people who are trying to use plugins with components
+  if (instanceFile.plugins) {
+    throw new Error(
+      'Serverless Framework plugins are not support by Serverless Components. Please remove the "plugins" property and try again'
```
medikoo, line 335: `are not supported`

src/cli/commands/utils.js
```diff
@@ -329,6 +329,13 @@ const loadVendorInstanceConfig = async (directoryPath, options = { disableCache:
     }
   }
 
+  // give early feedback to people who are trying to use plugins with components
+  if (instanceFile.plugins) {
+    throw new Error(
```
medikoo, line 334: Shouldn't this be reported as _operational_ error? I guess throwing it with bare `Error` will make reporting obscure (with full stack trace)

eahefnawy: what you mean "operational error"?! There's no stack trace by default. Here's how it looks like:

```
my-vue-app-2 node ../../../components/bin/bin deploy

Serverless › Serverless Framework Plugins are not support by Serverless Components. Please remove the "plugins" property and try again 

  Documentation: https://github.com/serverless/components 
  Support: https://app.serverless.com/support 
  Slack: https://www.serverless.com/slack/ 


my-vue-app-2 
```

src/cli/commands/utils.js
eahefnawy: good catch 👍

src/cli/commands/utils.js
medikoo: > what you mean "operational error"

See "Operational errors vs. programmer errors" section here: https://www.joyent.com/node-js/production/design/errors 
It neatly explains what I mean.

> There's no stack trace by default. Here's how it looks like:

Ok, so then _programmer_ errors which can be cryptic as "undefined is not a function" are also shown with no stack trace right? It can be painful for debugging

medikoo: In Framework we recognize between both error types, and _operational_ are exposed just by message, and _programmer_ errors always go with full stack trace.

(anyway that's unrelated to this PR)

medikoo (approved):

PR #793, opened 2020-09-01
chore(command): throw error for null aws credentials

### LINKED ISSUE #495: Throw warning if credentials object is empty
New AWS/Serverless users might not have AWS credentials setup in their machine, and they might not notice the `.env` documentations. They'd try to run a specific component, but AWS would return an unfriendly error.

For better DX, we should inform users that credentials are not found and they need to create a .env file.

AbderrahimSoubaiElidrissi:
## What has been implemented?

Throw an `Error` if credentials are not found  

Closes #495

## Steps to verify

- Run X
- ...

## Todos:

- [ ] Write tests
- [ ] Run Prettier

### DIFF
--- a/src/cli/commands/utils.js
+++ b/src/cli/commands/utils.js
@@ -157,8 +157,9 @@ const loadAwsCredentials = async () => {
   // get the credentials for that profile
   const credentials = parsedCredentialsFile[awsCredentialsProfile];
 
-  if (!credentials) return;
-
+  if (!credentials) {
+    throw new Error('Missing credentials. Please to create a .env file');
+  }
   // check if profile use assume role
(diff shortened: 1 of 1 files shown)


eahefnawy:
Thanks for jumping in @AbderrahimSoubaiElidrissi ... That's an old issue though and is no longer a concern. It should be closed. Sorry about that.

The way we're doing things right now is to let each component handle credentials errors based on their credentials needs, simply because component now could support any providers, not just AWS. So not having aws credentials no longer should throw an error.

AbderrahimSoubaiElidrissi:
Hi @eahefnawy  No worries I enjoyed doing this

PR #794, opened 2020-09-01
Techdebt cleanup for instance credentials

astuyve:
- Removes syntactic sugar of `async` around non-asynchronous functions
- Removes unused argument
- Removes superfluous call to loadInstanceCredentials, unneeded by the registry command, and only used elsewhere in runAll, which calls loadInstanceCredentials itself

### DIFF
--- a/src/cli/commands-cn/dev.js
+++ b/src/cli/commands-cn/dev.js
@@ -163,7 +163,7 @@ module.exports = async (config, cli) => {
   let instanceYaml = await utils.loadInstanceConfig(instanceDir);
 
   // Load Instance Credentials
-  const instanceCredentials = await utils.loadInstanceCredentials(instanceYaml.stage);
+  const instanceCredentials = utils.loadInstanceCredentials();
 
   const sdk = new ServerlessSDK({
     context: {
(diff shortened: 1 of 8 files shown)


eahefnawy:
@astuyve I think you've mixed up the global and china versions of the commands & utils. They're not 100% identical. The [global version is async](https://github.com/serverless/components/blob/master/src/cli/commands/utils.js#L216-L218), while the [china version is not](https://github.com/serverless/components/blob/master/src/cli/commands-cn/utils.js#L132). If we merge this PR, this would break the global version because **it's in fact async** and involves file operations and AWS API requests.

eahefnawy (changes requested):
@astuyve I think you've mixed up the global and china versions of the commands & utils. They're not 100% identical. The [global version is async](https://github.com/serverless/components/blob/master/src/cli/commands/utils.js#L216-L218), while the [china version is not](https://github.com/serverless/components/blob/master/src/cli/commands-cn/utils.js#L132). If we merge this PR, this would break the global version because it's in fact async and involves file operations and AWS API requests.

eahefnawy:
@astuyve I'm closing this. Thanks for jumping into it though 🙌

Issue #796, opened 2020-09-03
All variables in dotenv file are converted to string type


yugasun:
I found `serverless cli` using [dotenv](https://github.com/motdotla/dotenv) to parse `.env` file, but `dotenv` will convert all variables to `string`, if I config `number` type in `.env` for example `PORT=3306`, then my mysql connection will throw `port unknow error`.

Ralative issue: https://github.com/motdotla/dotenv/issues/51

medikoo:
@yugasun by design env variables can only be strings, you cannot set at CLI level a number value, it's also only strings that you can store at  `process.env` in Node.js (if you try to assign _number_, it'll be coerced to string).

Also recommended (and imo the only valid) use of `dotenv` is to at begin of program initialization transport all env vars from `.env` to `process.env`, and after that read them only from `process.env` (technically weather we're relying on `dotenv` or not, should be transparent to our app logic. It's the `process.env` that should be treated as the only source of truth)

In light of above, real fix for that issue, is to coerce `PORT` value to number before passing  it to mysql connection constructor.

medikoo:
Yes, technically you can store stringified JSON as env var, and then resolve it when needed.

I'm not sure exactly what problem we're trying to solve (?) Where is this mysql connection setup which raises issues?

yugasun:
@medikoo It's just an example, the key point is user want to keep the type of variables in .env config.

PR #797, opened 2020-09-04
fix: resolve content from env with strong type

zongUMR:
## What has been implemented?

Closes https://github.com/serverless/components/issues/796

## Steps to verify

This code will get content from `process.env` and convert them, return a new object with strong type.

`resolveVariables` will insert `number, string, boolean, null, undefined` content back to yaml when it has `${env:xxx}`

## Todos:

- [ ] Write tests
- [ ] Run Prettier


medikoo (commented):
@zongUMR what exactly use case we're trying to address here?

Technically by _nature_ `process.env` vars are strings, and it doesn't feel as a right move try to imply a generic mechanism to try handle that differently.

Ideally if specific typing is needed for some specific case, then best is to solve it when addressing given case specifically.

zongUMR:
@medikoo If an user input `export test=1 && sis deploy` and the config file has `${env:test}` here, then the old code will insert the `test` var is a string one, but user wants a number one, and when we send this var to the Tencent API, the API will validate the type of value, it needs a number as well, so we need to convert the real type of this value

medikoo:
> @medikoo If an user input export test=1 && sis deploy and the config file has ${env:test} here, then the old code will insert the test var is a string one. 

imo it's definitely an expected behavior.

If intention is to interpolate it to _number_ in some specific case, then I believe best if we introduce some type interpolation instructions at config var resolution level, e.g. in Framework we have such thing to e.g. interpolate booleans: https://www.serverless.com/framework/docs/providers/aws/guide/variables/#read-string-variable-values-as-boolean-values.

In components we may propose some simpler notation that will allow to interpolate to more types.

Note that trying to convert _string_ values unconditionally to some other type just by reasoning on their content may introduce a lot of issues, e.g. what if I unconditionally need a `"1"` or `"null"` string?

yugasun:
@medikoo Providing modify function for `${modifyFunc(env:xxx)}` is really a good method. Maybe we can add below function for most cases:

```
strToBool
strToNum
numToStr
```

medikoo:
@yugasun sounds good to me, @eahefnawy what do you think?

zongUMR:
That's good,  I will remove the `convert` code and only remain the `readme` changes.

PR #800, opened 2020-09-09
Supports passing inputs via the CLI

eahefnawy:
## What has been implemented?
Supports passing inputs via the CLI according to Option C in the following specs:

https://github.com/serverless/roadmap/issues/453
https://github.com/serverless/roadmap-tencent/issues/512

### DIFF
--- a/README.md
+++ b/README.md
@@ -650,6 +650,23 @@ Fetches information of an Instance of a Component.
 
 Starts DEV MODE, which watches the Component for changes, auto-deploys on changes, and (if supported by the Component) streams logs, errors and transactions to the terminal.
 
+#### `serverless <command> --inputs key=value foo=bar`
+
+Runs any component custom command, and passes inputs to it. The inputs you pass in the above syntax overwrite any inputs found in the `serverless.yml` file.
+
+Few examples:
+
+```
+# simple example
+serverless test --inputs domain=serverless.com
+
+# passing objects with JSON syntax
+serverless invoke --inputs env='{"LANG": "en"}'
+
+# passing arrays with comma seperation
+serverless backup --inputs userIds=foo,bar
+```
+
 <br/>
 
 # Building Components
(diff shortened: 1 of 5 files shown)


src/cli/utils.js
```diff
@@ -806,6 +806,61 @@ const validateNodeModules = async (directory) => {
...
+      return value.map((v) => setCliInputValueType(v));
+    }
+
+    // check if the string is actually a number
+    if (!isNaN(value)) {
+      value = Number(value);
+    }
+
+    // check if the string is actually boolean
+    value = value === 'true' || (value === 'false' ? false : value);
```
timqian, line 844: Will this be an issue when the user actually wants to pass the string "true"?

ac360:
Thx @eahefnawy.  We need object support in here, especially to pass event objects to serverless functions.  Perhaps we can try to auto-convert strings to objects, so users can use `event='{"foo": "bar"}'` like we do in V1?

src/cli/utils.js
eahefnawy: Yeah, it won't work. What's the use case for that though? For edge cases like this, I think users should just use the YAML file. There's only so much you can do with the CLI.

eahefnawy:
@ac360 hmm ok yeah that's better.

timqian:
I have tested this feature in China, `String`, `Array`, and `JSON` are all parsed correctly as described in the doc.

README.md
```diff
@@ -650,6 +650,23 @@ Fetches information of an Instance of a Component.
...
+
+```
+# simple example
+serverless test --inputs domain=serverless.com
+
+# passing objects with JSON syntax
+serverless invoke --inputs env='{"LANG": "en"}'
+
+# passing arrays with comma seperation
+serverless backup --inputs userIds=foo,bar
```
timqian, line 667: serverless component doesn't have  these methods: `invoke`, `test`, and `backup`, will it be better to avoid these in the examples?

eahefnawy: I don't think it matters. It's unlikely we'll find real-world examples that cover each of these use cases. I just want to demo how to use array, json...etc

timqian (approved):

PR #809, opened 2020-09-29
fix: customize command inputs

yugasun:
When running command is not `deploy`, we can not reset `inputs` to `{}`, because customized commands may need to use `inputs` in `serverless.yml`.

### DIFF
--- a/src/cli/commands-cn/utils.js
+++ b/src/cli/commands-cn/utils.js
@@ -224,7 +224,7 @@ const setInputsForCommand = (instanceYaml, command, config) => {
     const defaultInputs = command === 'deploy' ? instanceYaml.inputs : {};
     instanceYaml.inputs = instanceYaml.commandInputs[command] || defaultInputs;
   } else if (command !== 'deploy') {
-    instanceYaml.inputs = {};
+    // no op
   }
   // merging inputs from command args, e.g. slcc deploy --inputs.src="./new-src"
   // will be merged into inputs.src
--- a/src/cli/commands-cn/utils.js
+++ b/src/cli/commands-cn/utils.js
@@ -224,7 +224,8 @@ const setInputsForCommand = (instanceYaml, command, config) => {
     const defaultInputs = command === 'deploy' ? instanceYaml.inputs : {};
     instanceYaml.inputs = instanceYaml.commandInputs[command] || defaultInputs;
   } else if (command !== 'deploy') {
-    // no op
+    // make src undefined, so no need to upload code for other commands
+    instanceYaml.inputs.src = undefined
   }
   // merging inputs from command args, e.g. slcc deploy --inputs.src="./new-src"
   // will be merged into inputs.src
--- a/src/cli/commands-cn/utils.js
+++ b/src/cli/commands-cn/utils.js
@@ -225,7 +225,7 @@ const setInputsForCommand = (instanceYaml, command, config) => {
     instanceYaml.inputs = instanceYaml.commandInputs[command] || defaultInputs;
   } else if (command !== 'deploy') {
     // make src undefined, so no need to upload code for other commands
-    instanceYaml.inputs.src = undefined
+    instanceYaml.inputs.src = undefined;
   }
   // merging inputs from command args, e.g. slcc deploy --inputs.src="./new-src"
   // will be merged into inputs.src


src/cli/commands-cn/utils.js
```diff
@@ -224,7 +224,7 @@ const setInputsForCommand = (instanceYaml, command, config) => {
     const defaultInputs = command === 'deploy' ? instanceYaml.inputs : {};
     instanceYaml.inputs = instanceYaml.commandInputs[command] || defaultInputs;
   } else if (command !== 'deploy') {
-    instanceYaml.inputs = {};
```
timqian, line 227: This part looks intentional. @hkbarton Why are we resetting `inputs` to `{}` here?

timqian:
@yugasun What are the `customized commands` you mentioned that will make use of the `inputs` in `serverless.yml`.

yugasun:
@timqian For cynosdb component, use may need reset theire password, or some other configs, they just config it in `serverless.yml` and using `sls resetpwd` command.

src/cli/commands-cn/utils.js
yugasun: Maybe no need to reset `inputs` to `{}`, so customize command can use these `inputs` directly.

yugasun:
@timqian I guess @hkbarton resetting `inputs` to `{}` just want to prevent source code to upload by `inputs.src`, so I make `inputs.src` to `undefined`.

src/cli/commands-cn/utils.js
```diff
@@ -224,7 +224,8 @@ const setInputsForCommand = (instanceYaml, command, config) => {
...
+    // make src undefined, so no need to upload code for other commands
+    instanceYaml.inputs.src = undefined;
```
hkbarton, line 228: maybe let's do this:

```
if (!instanceYaml.inputs) {
   instanceYaml.inputs = {}
}
```

I don't think we should remove src for user, if in case use DO need put src in their commands' input. 

I think the problem here is we should check if inputs exist or not, and we should NOT rest it to `{}` if the `inputs` is given.

hkbarton (changes requested):
see comment inline

src/cli/commands-cn/utils.js
hkbarton: by doing this we are simply sanitizing the inputs, not changing any data inside.

yugasun: But when run commands except `deploy` for example `remove`, if we pass `inputs.src` to it, this action will also do `upload task`, below is the relative code `src/instance.js` in npm module `@serverless/platform-client-china`:

```js
const run = async (sdk, method = {}, instanceData = {}, credentials = {}, options = {}) => {
  let instance = JSON.parse(JSON.stringify(instanceData));

  // Validate method
  if (!method) {
    throw new Error('A "method" argument is required');
  }

  // Validate instance
  instance = validateAndFormat(instance);

  // Run source hook and upload source, if "src" input is used...
  // Save original src to present neatly in Dashboard/clients, or this will be replaced by the S3 location by the functions below
  if (instance.inputs.src) {
    instance.inputs.srcOriginal = JSON.parse(JSON.stringify(instance.inputs.src));
  }
  instance.inputs.src = await preRunSrcHook(instance.inputs.src);

  let size;
  let uploadTask;
  if (instance.inputs.src && typeof instance.inputs.src === 'string') {
    // Process source code, or skip if not provided...
    const { src: source, uploadPromise: task, cosInfo: sourceCosInfo } = await preRunSrcUpload(
      sdk,
      instance.orgName,
      instance.inputs,
      options
    );
    if (options.debug) {
      console.debug(`the code will be uploaded to ${source}`);
    }
    instance.inputs.src = source;
    instance.inputs.sourceCosInfo = sourceCosInfo;
    uploadTask = task;
  }

  if (uploadTask) {
    await uploadTask;
    await api.instance.run(sdk, instance, method, credentials, options, size);
  } else {
    return api.instance.run(sdk, instance, method, credentials, options, size);
  }
  return null;
};
```

hkbarton: @zongUMR has put a fix in this PR https://github.com/serverlessinc/platform-cn/pull/107, we think the fix belongs to here

hkbarton: basically doing:
```
  if (instance.inputs.src && typeof instance.inputs.src === 'string' && method === 'deploy')  
```
in the code you pasted

hkbarton: and the rule of here is don't mutate user's inputs

yugasun: Well, if your [serverlessinc/platform-cn#107](https://github.com/serverlessinc/platform-cn/pull/107) have been merged, I will close this PR.

PR #813, opened 2020-10-01
Catch a 401 on the backend and handle the error on the CLI

### LINKED ISSUE #812: SLS commands crash hard when the org doesn't exist
# Description
Looks like we just need to wrap this ws connection in a try/catch block, and interpret if we get a 404 from the backend API call:
```
SERVERLESS_PLATFORM_STAGE=dev sls deploy --debug
Initializing...
Action: "deploy" - Stage: "dev" - Org: "ajstuyvenbergasasdfasdf" - App: "js-mn" - Name: "js-mn"
Running in Platform Dev stage

 Error: Unexpected Serverless Platform Streaming Error: Unexpected server response: 401
    at WebSocket.self.connection.onerror (/Users/aj/serverless/platform/sdk/src/Connection.js:95:11)
    at WebSocket.onError (/Users/aj/serverless/platform/sdk/node_modules/ws/lib/event-target.js:133:16)
    at WebSocket.emit (events.js:203:13)
    at abortHandshake (/Users/aj/serverless/platform/sdk/node_modules/ws/lib/websocket.js:697:15)
    at ClientRequest.<anonymous> (/Users/aj/serverless/platform/sdk/node_modules/ws/lib/websocket.js:579:7)
    at ClientRequest.emit (events.js:203:13)
    at HTTPParser.parserOnIncomingClient [as onIncoming] (_http_client.js:565:23)
    at HTTPParser.parserOnHeadersComplete (_http_common.js:116:17)
    at TLSSocket.socketOnData (_http_client.js:452:22)
    at TLSSocket.emit (events.js:203:13)

Serverless › Unexpected Serverless Platform Streaming Error: Unexpected server response: 401

  Documentation: https://github.com/serverless/components
  Support: https://app.serverless.com/support
  Slack: https://www.serverless.com/slack/
```

astuyve:
## What has been implemented?

Closes #812

## Steps to verify

- Run `sls deploy`, `sls dev`, or other commands with an organization that does not exist
- [ ] You see a helpful error message, instead of `internal server error`

image

There may be a better way to do this, and I'm totally open to suggestions - but I believe our websockets implementation only provides a string error.

Alternatively we could refactor the SDK to throw errors with status codes, and then inspect them here.

### DIFF
--- a/src/cli/commands/dev.js
+++ b/src/cli/commands/dev.js
@@ -230,6 +230,13 @@ module.exports = async (config, cli) => {
       onEvent,
     });
   } catch (error) {
+    if (error.message.includes('401')) {
+      cli.sessionStop(
+        'error',
+        `Your credentials do not have access to the Organization with the name of: ${instanceYaml.org}.  Try logging into a different account.`
+      );
+      return null;
+    }
     throw new Error(error);
   }
 
--- a/src/cli/commands/run.js
+++ b/src/cli/commands/run.js
@@ -38,7 +38,6 @@ module.exports = async (config, cli, command) => {
 
   // Get access key
   const accessKey = await getAccessKey(instanceYaml.org);
-
   const meta = `Action: "${command}" - Stage: "${instanceYaml.stage}" - Org: "${instanceYaml.org}" - App: "${instanceYaml.app}" - Name: "${instanceYaml.name}"`;
 
   // Presentation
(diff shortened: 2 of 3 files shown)


eahefnawy (approved):
Looks good! I think these status codes would be better handled in the SDK though.

We are relying on that error message, which we may easily change. It also only handles 401, while the linked issue describes 404, so we may face this issue again.

Nonetheless, I think we should merge this change as its better than not having any error handling at all.

Thanks @astuyve 🙌

astuyve:
Yup! Will refactor this once we release a new major rev of the SDK that includes error objects w/ codes

PR #818, opened 2020-10-12
fix: parse cli --input args

zongUMR:
According this PR: https://github.com/serverless/components/pull/800 and https://github.com/serverless/components/pull/809, we only need to write `commandInputs` and insert it into final config

### DIFF
--- a/src/cli/commands-cn/utils.js
+++ b/src/cli/commands-cn/utils.js
@@ -258,6 +258,12 @@ const saveYaml = async (yamlPath, yamlObj) => {
   const yamlContent = YAML.safeDump(yamlObj);
   await fse.writeFile(yamlPath, yamlContent);
 };
+const setInputsForCommandInputs = (instanceYaml, command) => {
+  // If user sets customized command inputs in yaml, need to insert them in final yaml config
+  if (instanceYaml.commandInputs && instanceYaml.commandInputs[command]) {
+    instanceYaml.inputs = { ...instanceYaml.inputs, ...instanceYaml.commandInputs[command] };
+  }
+};
 
 module.exports = {
   loadInstanceConfig: loadTencentInstanceConfig,
@@ -269,4 +275,5 @@ module.exports = {
   handleDebugLogMessage,
   parseYaml,
   saveYaml,
+  setInputsForCommandInputs,
 };
--- a/src/cli/commands-cn/utils.js
+++ b/src/cli/commands-cn/utils.js
@@ -219,20 +219,6 @@ const getInstanceDashboardUrl = (instanceYaml) => {
   return `Full details: https://serverless.cloud.tencent.com/apps/${instanceYaml.app}/${instanceYaml.name}/${instanceYaml.stage}`;
 };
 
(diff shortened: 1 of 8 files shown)


hkbarton (dismissed):

src/cli/commands-cn/utils.js
```diff
@@ -228,8 +228,10 @@ const setInputsForCommand = (instanceYaml, command, config) => {
   }
   // merging inputs from command args, e.g. slcc deploy --inputs.src="./new-src"
   // will be merged into inputs.src
-  if (config.inputs) {
-    instanceYaml.inputs = mergeDeepRight(instanceYaml.inputs, config.inputs);
+
+  const cliInputs = parseCliInputs();
+  if (cliInputs) {
+    instanceYaml.inputs = mergeDeepRight(instanceYaml.inputs, cliInputs);
```
timqian, line 234: The `cliInputs` seems has been merged to `instanceYaml.inputs` here: https://github.com/serverless/components/pull/818/files#diff-13feadcc91559c6a62f296a753afb4afL33
Why do we need to do it here again?

zongUMR: The above code will overwrite the inputs, so it will insert the CLI args

src/cli/commands-cn/utils.js
```diff
@@ -219,7 +219,7 @@ const getInstanceDashboardUrl = (instanceYaml) => {
   return `Full details: https://serverless.cloud.tencent.com/apps/${instanceYaml.app}/${instanceYaml.name}/${instanceYaml.stage}`;
 };
 
-const setInputsForCommand = (instanceYaml, command, config) => {
+const setInputsForCommand = (instanceYaml, command) => {
```
timqian, line 222: It should be safe for us to remove this function, it is not invoked elsewhere

timqian (approved):

zongUMR:
@yugasun You can review it and give any feedback. thanks

src/cli/commands-cn/utils.js
```diff
@@ -272,6 +258,12 @@ const saveYaml = async (yamlPath, yamlObj) => {
   const yamlContent = YAML.safeDump(yamlObj);
   await fse.writeFile(yamlPath, yamlContent);
 };
+const setInputsForCommandInputs = (instanceYaml, command) => {
+  // If user sets customized command inputs in yaml, need to insert them in final yaml config
+  if (instanceYaml.commandInputs && instanceYaml.commandInputs[command]) {
+    instanceYaml.inputs = { ...instanceYaml.inputs, ...instanceYaml.commandInputs[command] };
```
timqian, line 264: In this way inputs from `commandInputs` of `serverless.yml` have higher priority than inputs from the command line (`sls action --input`). Is this what we want?

zongUMR: let me confirm that

zongUMR: fix it, thanks

timqian (approved):

PR #824, opened 2020-10-21
feat: add origin src to inputs

zongUMR:
`src` will be added `absolute path`, I add another field to save the original path(src and dist), so we can show the current info on website:
Screen Shot 2020-10-21 at 10 17 00

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [3.2.5](https://github.com/serverless/components/compare/v3.2.4...v3.2.5) (2020-10-21)
+
+- Feat: Add original src in inputs
+
 ## [3.2.4](https://github.com/serverless/components/compare/v3.2.2...v3.2.4) (2020-10-17)
 
 - Feat: Add a traceId into proxy for tracing a full-chain request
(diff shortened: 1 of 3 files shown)


zongUMR:
@yugasun Please review this PR

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [3.2.5](https://github.com/serverless/components/compare/v3.2.4...v3.2.5) (2020-10-21)
+
+- Feat: Add original src in inputs
```
timqian, line 7: This feature has not been released as well, maybe add it here as well? https://github.com/serverless/components/pull/823

yugasun:
Well done, I have another question, when using `getInstance` to get instance detail, `inputs.src` of detail will be revert to `originSrc` or both `src` and `originSrc` will exist?

CHANGELOG.md
zongUMR: Your meaning is that part of code was merge into `master` but do not want to release?

zongUMR:
> Well done, I have another question, when using `getInstance` to get instance detail, `inputs.src` of detail will be restored to `originSrc` or both `src` and `originSrc` will exist?

They will be both saved

CHANGELOG.md
timqian: Yes~

timqian (approved):

yugasun:
So when web page try to display the result of deployment, the `src` will still be absolute path?

zongUMR:
> So when web page try to display the result of deployment, the `src` will still be absolute path?

Yes, I do not change the old logic for `src` to avoid any problem from this field, just add a new one

yugasun:
Well, maybe the frontend can try to restore `src` to `originSrc`.

PR #825, opened 2020-10-22
use English description as fallback if no zh-cn

timqian:


src/cli/interactive-onboarding/cn.js
```diff
@@ -79,14 +79,14 @@ const getTemplatesFromRegistry = async (sdk) => {
...
     .map((item) => ({
-      name: `${item.name} - ${item['description-i18n']['zh-cn']}`,
+      name: `${item.name} - ${item['description-i18n'] ? item['description-i18n']['zh-cn'] : item['description']}`,
       value: { id: item.componentName, name: item.name },
     }));
 
   const scfTemplatesChoices = templates
     .filter((item) => item.name.includes('scf-') && item.name !== 'scf-starter')
     .map((item) => ({
-      name: `${item.name} - ${item['description-i18n']['zh-cn']}`,
```
zongUMR, line 89: How to show the message if `item['description-i18n']` exist but no `item['description-i18n']['zh-cn']`?

timqian: It will show `nodejs-starter - undefined`.
Yeah, we can do better, just updated this logic

zongUMR (approved):

PR #829, opened 2020-10-28
Minor fixes for publish, init and deploy

eahefnawy:
- Remove `app` property requirement from `sls publish` (because the app is determined on `init`)
- Remove `name` and `stage` properties from root level `serverless.yml`
- Remove inputs from the yaml file when running `serverless remove`
- Remove support for single command deploys.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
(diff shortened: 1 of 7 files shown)


medikoo (approved):
When we will see some tests setup for this project? It'll be great to have a mean to confirm that such changes aren't breaking in some way, automatically

eahefnawy:
I think @ac360 mentioned that we want to rely more on integration tests for a more accurate picture of what might be broken, which I think we do. We might just want to add tests for these new cases.

medikoo:
We have some tests in platform and they do not even run in CI/CD.

My point was to have some testing automated, so we automatically confirm before merging that we do not introduce a regression, like we do in a Framework, and technically like all other popular production ready products do

PR #836, opened 2020-11-05
Generate serverless.yml automatically for web components

timqian:
When users run sls deploy in a directory without a serverless.yml, check if it is possible to transform this project into a serverless project and provide a friendly guidance

Related issue: serverless/roadmap-tencent#603


src/cli/index.js
```diff
@@ -13,11 +13,11 @@ const semver = require('semver');
 const chalk = require('chalk');
 const HttpsProxyAgent = require('https-proxy-agent');
 const CLI = require('./CLI');
-const { loadInstanceConfig, fileExistsSync, isProjectPath, isChinaUser } = require('./utils');
+const { loadInstanceConfigUncached, fileExistsSync, isProjectPath, isChinaUser } = require('./utils');
 
 module.exports = async () => {
   const args = minimist(process.argv.slice(2));
-  const instanceConfig = loadInstanceConfig(process.cwd());
+  const instanceConfig = loadInstanceConfigUncached(process.cwd());
```
timqian, line 20: Have to uncached `instanceConfig` as we will need to auto-generate it if it does not exist.

We use `memoizeWith` of `ramda` and it doesn't support clear `memoizeWith` items 
 
https://github.com/ramda/ramda/issues/2509

src/cli/commands-cn/run.js
```diff
@@ -21,6 +22,16 @@ module.exports = async (config, cli, command) => {
     return runAll(config, cli, command);
   }
 
+  if (
+    command === 'deploy' &&
+    !fs.existsSync(path.join(process.cwd(), 'serverless.yml')) &&
```
zongUMR, line 27: Will it check user's config if the project has a `Serverless.yaml`? You can check the possible config file in https://github.com/serverless/components/blob/f945ae9ce02831702538af129bf3e554b7b911b2/src/cli/utils.js#L443

timqian: Thanks for pointing it out, switched to this method
https://github.com/serverless/components/blob/feature-yml_auto_generation/src/cli/serverlessFile.js#L16

zongUMR (approved):

src/cli/commands-cn/run.js
```diff
@@ -15,12 +16,23 @@ const generateNotificationsPayload = require('../notifications/generate-payload'
...
     return runAll(config, cli, command);
   }
 
+  if (
+    command === 'deploy' &&
+    !getServerlessFilePath(process.cwd()) &&
+    fs.existsSync(path.join(process.cwd(), 'package.json'))
+  ) {
+    const generatedYML = await utils.generateYMLForNodejsProject();
+    fs.writeFileSync(path.join(process.cwd(), 'serverless.yml'), generatedYML, 'utf8');
```
medikoo, line 32: Let's use async variant

src/cli/commands-cn/utils.js
```diff
@@ -269,6 +270,164 @@ const saveYaml = async (yamlPath, yamlObj) => {
...
+  runtime: Nodejs10.15
+  apigatewayConf:
+    protocols:
+      - http
+      - https
+    environment: release
+`;
+
+  const supportedComponents = ['express', 'koa', 'egg', 'next', 'nuxt'];
+  const packageJsonFile = fs.readFileSync(path.join(process.cwd(), 'package.json'), 'utf-8');
```
medikoo, line 354: Let's use async variant

src/cli/commands-cn/utils.js
```diff
@@ -269,6 +270,164 @@ const saveYaml = async (yamlPath, yamlObj) => {
...
+      message: '在 package.json 里发现以下依赖，选择您希望创建的 serverless 的应用类型',
+      type: 'list',
+      name: 'ymlType',
+      choices: knownPackages,
+    });
+    ymlType = result.ymlType;
+  }
+
+  if (ymlType === 'express') {
+    if (fs.existsSync(path.join(process.cwd(), 'sls.js'))) {
```
medikoo, line 383: Let's use async variant (and in other places)

src/cli/commands-cn/run.js
```diff
@@ -15,12 +16,23 @@ const generateNotificationsPayload = require('../notifications/generate-payload'
...
   if (!config.target && runningTemplate(process.cwd())) {
     return runAll(config, cli, command);
   }
 
+  if (
+    command === 'deploy' &&
+    !getServerlessFilePath(process.cwd()) &&
+    fs.existsSync(path.join(process.cwd(), 'package.json'))
+  ) {
+    const generatedYML = await utils.generateYMLForNodejsProject();
```
medikoo, line 31: If I see correctly, it'll trigger interactive setup when user does `sls deploy`. Is this intended?
What if `sls deploy` is triggered in CI or other automated environment?

Is it a good idea to mix interactive setup into a deployment command?

src/legacy.js
```diff
@@ -52,6 +52,10 @@ const runningComponents = () => {
     return process.argv.length === 2 && isChinaUser();
   }
 
+  if (!componentConfig && process.argv[2] === 'deploy' && isChinaUser()) {
+    return true;
```
medikoo, line 56: This feels slightly problematic, as it'll also be triggered for scenarios when:
- User runs `sls deploy` in context of a V1 service with configuration that has some error
- User runs `sls deploy` not in service context, simply by mistake

src/cli/index.js
medikoo: If we do not rely on caching here, then I think this cached version has no longer any use.

We can switch to `memoizee` which allows to clear cache (and it's used by other dependencies, so it won't add to package size)

src/cli/commands-cn/run.js
timqian: This is intended, we discussed in the meeting last week
https://github.com/serverless/roadmap-tencent/issues/603
The idea is to `deploy a project without a serverless.yml`, so this process is added in the `deploy` command

src/cli/commands-cn/run.js
timqian: As we only have one thread here, I guess `fs.promises.writeFile` and `fs.writeFileSync` is the same?

src/cli/index.js
timqian: Thanks, let me check memoizee out

src/cli/commands-cn/run.js
medikoo: Ok, thanks for clarification

src/cli/commands-cn/run.js
medikoo: Not really. JS is in general one thread, but this single thread can process other things while waiting for IO.
It's the whole idea of asynchronous programming, that instead waiting for IO you let the process to handle other things in a meantime 

In our case we sometimes to other work in a background, like sending analytics report

src/cli/commands-cn/run.js
medikoo: Why do we envision projects without `serverless.yml` in its directory  in first place?

Is it because we first supported such model, and now we want to ensure that all projects have `serverless.yml` at the root?

src/legacy.js
timqian: > User runs sls deploy in context of a V1 service with a configuration that has some error
Thanks, I have excluded V1 service as well

> User runs sls deploy not in service context
This is intended, we want to transform a non-serverless project to a serverless project. And the users will be able if they want to

src/cli/commands-cn/run.js
timqian: It is for users to deploy existed web servers (for example a normal express APP).

I tested once before and thought we can directly deploy normal express/koa code after the serverless.yml is generated. But after doing more tests and look into the `tencent-componet` code, I find users can not directly deploy normal express/koa code, they will need to export the app. In this way, I think auto-generating `serverless.yml` may not solve the problem they want to solve. I will discuss this with the Tencent team again.

timqian: Update
> We decided to keep moving forward with this and prompt users on the initial deployment (if there is no serverless.yml) to remind them they have to export the app for koa and express.

https://github.com/serverless/roadmap-tencent/issues/603#issuecomment-728674394

timqian:
Updates
- [x] Use [momoizee]() to cache and remove the cache after `serverless.yml` is changed
- [x] Use async variant when reading/writing files
- [x] Remind users that they have to export the app for koa and express apps

src/cli/index.js
timqian: Switched to memoizee

src/cli/commands-cn/run.js
timqian: Switched to async variant

zongUMR (dismissed):

src/cli/commands-cn/utils.js
```diff
@@ -269,6 +270,144 @@ const saveYaml = async (yamlPath, yamlObj) => {
   await fse.writeFile(yamlPath, yamlContent);
 };
 
+const generateYMLForNodejsProject = async (cli) => {
```
yugasun, line 273: Maybe we can put these template functions in an alone js file.

src/cli/commands-cn/utils.js
```diff
@@ -269,6 +270,144 @@ const saveYaml = async (yamlPath, yamlObj) => {
...
+  const getExpressYML = (entryFile) => `component: express
```
yugasun, line 274: Maybe we can put these template functions in an alone js file.

timqian (commented):
I don't see a good place to store the extracted function. Guess it is our convention to store all utils related functions in this file.

src/cli/commands-cn/utils.js
timqian: I don't see a good place to store the extracted function. Guess it is our convention to store all utils related functions in this file.

yugasun: Well, just ignore my advise.

src/cli/commands-cn/run.js
medikoo: Why user would run `sls deploy` in typical express app that has no serverless configuration? Are we suggesting it in our docs?

src/legacy.js
medikoo: If I see correctly it'll still resolve to true for Framework user, where service config has YAML error, or AWS variables are used (which is not common).

I think, for it to be correct, we need to check whether any kind of serverless config file (in whatever shape, note that e.g. we also support `serverless.ts`)) is not present.

src/cli/commands-cn/run.js
timqian: Yes, we plan to suggest it in corresponding components' README after this feature is released

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@serverless/components",
-  "version": "3.3.1",
+  "version": "3.3.2",
```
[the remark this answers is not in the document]
timqian: Thanks for the note, the version is updated

src/legacy.js
timqian: Thanks, checking existance of serverless config files here instead

medikoo (approved):

PR #840, opened 2020-11-18
Do not automatically load credentials when deploying the express component

eahefnawy:
Only allows the user to explicitly set credentials with `.env` file, or providers.


medikoo (commented):
@eahefnawy I see it's marked as WIP. Is this not finished? If it is, was it intentionally passed for review?

(Also GitHub has now concept of "Draft" PR's, I think they serve as great mean for marking WIP work)

PR #842, opened 2020-11-20
sls info command for nested templates

timqian:
Enable `sls info` and `sls info --debug` command for nested templates

### Related Issue

https://github.com/serverless/roadmap-tencent/issues/621


eahefnawy (changes requested):
Looks good @timqian ... this PR doesn't seem to support `sls info --debug`, which shows the state of the instance when running in a single instance. Don't we also want that when running nested templates? For consistency at least.

timqian:
@eahefnawy Thanks for the suggestion, I will add `--debug` support

eahefnawy (approved):
Looks good Tim! 🙌

PR #863, opened 2021-01-06
Refactor credentials handling

eahefnawy:
## What has been implemented?

- credentials will only be loaded from .env files
- credentials will no longer be automatically loaded from the environment
- credentials will no longer be automatically loaded from the aws credentials file in the home directory
- only if the user explicitly specified AWS_PROFILE env var in the .env file, only then we will load this profile credentials from the aws credentials file
- also fixes a bug that loaded the wrong file .env for the given stage

**Note: This is a necessary breaking change**

## Steps to verify

Deploy a component without credentials in the .env file, but with credentials in the environment or home directory. The deployment should fail.

PR #867, opened 2021-01-07
Fix load order of env files in second level parent directories

eahefnawy:


eahefnawy:
I'll self-approve this simple change to get it out quickly and get the bug fixed. It's already been approved in #863.

PR #878, opened 2021-01-28
Do not load credentials from ~/.aws/credentials

eahefnawy:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@serverless/components",
-  "version": "3.5.1",
+  "version": "3.6.0",
```
eahefnawy, line 3: While this is a breaking change, bumping a minor version will make sure this is released asap. Otherwise, we'll have to also manually make a framework release using the major version. Imo, using the major version in this case is useless cause users are still not able to pin to a specific components version.

PR #879, opened 2021-01-31
BUG FIX - dev mode does not watch parent directories

### LINKED ISSUE #871: Dev Mode doesn't work when "src" points to parent directories
I have a use-case where I need to share custom code across two different apps.  Because AWS Lambda Layers requires publishing to npm, I'd prefer not to use that to avoid the npm overhead.  Because Components does not have `include` functionality in the `src` input, I can't reference a local folder w/ shared code.

To solve this, I've put a `code` folder in a parent directory, and I have two `serverless.yml` files in separate sub folders, with their `src` pointing to `code`.  This works, but `serverless dev` mode does not work.  Specifically the folder watching / re-deploy part of dev mode.

eahefnawy:
## What has been implemented?
Closes #871 

Fixes a bug where dev mode does not watch parent directories.

## Steps to verify

Run dev mode on an instance with a src input pointing to a parent directory:

```yml
component: express
name: my-app

inputs:
  src: ../src
```

Then make changes to the src files. Without this PR changes, a redeploy will not be triggered. With this PR changes, a redeploy will be triggered.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [3.6.1](https://github.com/serverless/components/compare/v3.6.0...v3.6.1) (2021-01-28)
+
+- Fix a bug when using dev mode and the src input is pointing to a parent directory
+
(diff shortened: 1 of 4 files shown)


ac360 (approved):
I am unable to do a proper code review at this time.  If you think this will suffice then pls move forward @eahefnawy, or reach out to someone else.

dougmoscrop (approved):

PR #880, opened 2021-01-31
Do not load local credentials

eahefnawy:
This PR removes support for loading credentials from the user machine, specifically from `.env` files, and instead relies on providers, which are loaded via the platform backend.


harrikauhanen:
This caused confusion to anyone using local credentials. I think things should not just suddenly stop working (unless of course you e.g. update your local packages)  See: https://serverless-contrib.slack.com/archives/CA4QT5VU3/p1613426692259600

PR #881, opened 2021-02-01
refactor: Use `utils` instead of `platform-sdk` in `login`

pgrzesik:
## What has been implemented?

Addresses parts of: https://github.com/serverless/enterprise-plugin/issues/464

- replace `writeConfigFile` and `readConfigFile` + config-related logic with `configUtils` equivalents
- replace `platform-sdk`'s `listTenants` with `listOrgs`
- replace `platform-sdk`'s `refreshToken` with `utils` version of `refreshToken`
- remove dead code `getOrCreateAccessKey` - I've search for that in all projects in `serverless` and `serverlessinc` orgs and I didn't find anything that uses it

To make the review process easier, I've implemented all self-contained changes as separate commits. Unfortunatelly, there are not automated tests, so everything was verified by hand by running `login`, `logout`, `deploy` and `publish/unpublish` commands. 

## Steps to verify

In order to verify, I've run the following commands:

- Run `serverless login`
- Run `serverless logout`
- Run `serverless publish` with a new component and remove default org from `.serverlessrc` (so it is actually fetched via SDK call)

After each step, assess that the content of the `.serverlessrc` file looks like expected (functionally it should work the same as previously).

### DIFF
--- a/src/cli/commands/utils.js
+++ b/src/cli/commands/utils.js
@@ -6,7 +6,6 @@
 
 const args = require('minimist')(process.argv.slice(2));
 const path = require('path');
-const { listTenants } = require('@serverless/platform-sdk');
 const { ServerlessSDK } = require('@serverless/platform-client');
 const configUtils = require('@serverless/utils/config');
 const accountUtils = require('@serverless/utils/account');
@@ -52,10 +51,12 @@ const getDefaultOrgName = async () => {
 
   // if defaultOrgName is not in RC file, fetch it from the platform
   if (!defaultOrgName) {
-    await accountUtils.refreshToken(new ServerlessSDK());
+    const sdk = new ServerlessSDK();
+    await accountUtils.refreshToken(sdk);
 
     const { username, idToken, userId } = configUtils.getLoggedInUser();
-    const orgsList = await listTenants({ username, idToken });
+    sdk.config({ accessKey: idToken });
+    const orgsList = await sdk.listOrgs(username);
 
     // filter by owner
     const filteredOrgsList = orgsList.filter((org) => org.role === 'owner');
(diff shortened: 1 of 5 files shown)


src/cli/commands/utils.js
```diff
@@ -55,22 +51,18 @@ const getDefaultOrgName = async () => {
...
-    await refreshToken();
-
-    const userConfigFile = readConfigFile();
+    await accountUtils.refreshToken(new ServerlessSDK());
 
-    const { username, dashboard } = userConfigFile.users[userConfigFile.userId];
-    const { idToken } = dashboard;
-    const orgsList = await listTenants({ username, idToken });
+    const { username, idToken, userId } = configUtils.getLoggedInUser();
+    const sdk = new ServerlessSDK({ accessKey: idToken });
```
medikoo, line 57: It looks a bit dirty, that we initialize `ServerlessSDK` twice. Is there no way to stick to first instance?

medikoo (commented):
@pgrzesik Looks great to me! I have just one API usage concern

src/cli/commands/utils.js
pgrzesik: Good catch - updated

medikoo (dismissed):
Looks great to me @pgrzesik (still let's leave final approval to @eahefnawy )

eahefnawy (approved):
Looks great @pgrzesik ... super happy we've **finally** doing this!

Not sure if you wanna publish this right away, in that case, you'll need to bump the sdk version and add changelogs.

pgrzesik:
Updated version & changelog - could you take one last look @eahefnawy?

eahefnawy (approved):
Good job @pgrzesik 👍

PR #883, opened 2021-02-03
Improve Chinese users' experience

timqian:
> Related issue https://github.com/serverless/roadmap-tencent/issues/718

## What has been implemented?

- [x] ~~Make sure Chinese user always use components~~
- [x] ~~Make sure Chinese user see the right commands when `sls --help`~~
- [x] Make sure the user is in a serverless project before asking him/her to login in


zongUMR (approved):

src/legacy.js
```diff
@@ -35,28 +26,18 @@ const runningComponents = () => {
     return true;
   }
 
-  // Chinese users running "serverless deploy" in a project without a serverless config file
-  if (isChinaUser() && process.argv[2] === 'deploy' && !hasServerlessConfigFile(process.cwd())) {
+  // Load componets if user in China
+  if (isChinaUser()) {
```
medikoo, line 30: If I see correctly it'll create regression for traditional framework users in china, as after this update, Framework will stop working.

Is that correct?

medikoo (dismissed):

src/legacy.js
timqian: Yes, we should take care of that.

But we still want to let the Chinese users see the docs for components when they execute `sls --help` in China. I will just check if users input that command.

timqian: @ole3021 @ac360 @zongUMR 
In the Chinese documentation on the `serverless` project (https://github.com/serverless/serverless/blob/master/README_CN.md). I find it is describing the earlier version of serverless but the example can not be deployed successfully.

Do you know if we still support that kind of usage?

ole3021: Will check what the functions do and if we can support these features.

medikoo: @timqian Framework will show help for components if `--help-components` is passed, and it's also indicated in `--help` output:

Screenshot 2021-02-04 at 15 17 09

medikoo: @timqian Still I agree that it can be lost, what we can do.. is to add a notice in chinese about `--help-components` in first line of `--help` output, when command is run in China

ole3021:
@timqian  I did tests about the `serverless-tencent-scf` and it works, the issue you have above is because you didn't do `npm install` after init the project. which means

- [ ] we need to double-check how the SCF usages status on Tencent Cloud with Serverless. 
- [ ] What features are supported and what are not

I think we can't simple use the component help info relace all the sls help info, as the sls function is the main product feature and more important than the component.

zongUMR (approved):

ole3021 (approved):

medikoo (commented):
@timqian does this change make any difference?

All the change I see, is that _login_ now happens after we read service yaml and not before

timqian:
@medikoo If the folder does not contain a valid serverless YAML, we won't need to ask the user to login(scan QR code), but throws an error beforehand

PR #897, opened 2021-02-25
chore: Drop `@serverless/platform-sdk` dependency

pgrzesik:
## What has been implemented?

Removes dependency on `urls` as passing `loginBrokerUrl` explicitly to `platform-client`'s login is no longer needed. In turn, that allows us to drop `@serverless/platform-sdk` altogether.'

I tested it locally with `login` and `logout` commands.

## Steps to verify

- Manually login & logout

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,9 +33,8 @@
     "/templates"
   ],
   "dependencies": {
-    "@serverless/platform-client": "^4.0.0",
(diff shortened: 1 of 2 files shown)


medikoo (approved):
Great to see `@serverless/platform-sdk` being dropped here!

eahefnawy (approved):
Wooo! Amazing! 🙌

PR #898, opened 2021-03-01
fix: check invalid inline comments in env file

zongUMR:
## What has been implemented?

Closes https://github.com/serverless/roadmap-tencent/issues/787

Currently, the `dotenv` can not remove **# inline comment** in the env file: https://github.com/motdotla/dotenv/issues/484, which will cause error when users use inline comments. This PR will check this situation and throw an message.

**Question**: Is it possible that the value in the `env file` itself contains the # symbol?

### DIFF
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -107,19 +107,24 @@ module.exports = async () => {
 
   // Check env file whether or not containing an inline comment, which is invalid for the dotenv package's parsing: https://github.com/motdotla/dotenv/issues/484
   if (dotEnvContent && dotEnvContent.parsed) {
+    const regexForComment = / #+.*$/g;
     const envKeys = Object.keys(dotEnvContent.parsed);
     envKeys.forEach((key) => {
       const envValue = dotEnvContent.parsed[key];
-      if (envValue.includes('#')) {
+      if (regexForComment.test(envValue)) {
         if (isChinaUser()) {
-          cli.log(`${key}的值为:${envValue}, 含有不正确内容，请修改后重试(不支持行内注释)`, 'red');
+          cli.log(
+            chalk.yellow(
+              `在dotenv配置中字段${key}发现 #,请确保注释都写在单独由#开头的新一行, 不支持行内注释。详情查看：https://github.com/motdotla/dotenv#rules`
+            )
+          );
         } else {
           cli.log(
-            `The value of ${key} is: ${envValue}, which contains invalid content, please retry it after changing(No support for inline comment)`,
-            'red'
+            chalk.yellow(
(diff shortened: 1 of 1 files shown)


src/cli/index.js
```diff
@@ -105,6 +105,25 @@ module.exports = async () => {
...
   }
 
+  // Check env file whether or not containing an inline comment, which is invalid for the dotenv package's parsing: https://github.com/motdotla/dotenv/issues/484
+  if (dotEnvContent && dotEnvContent.parsed) {
+    const envKeys = Object.keys(dotEnvContent.parsed);
+    envKeys.forEach((key) => {
+      const envValue = dotEnvContent.parsed[key];
+      if (envValue.includes('#')) {
+        if (isChinaUser()) {
+          cli.log(`${key}的值为:${envValue}, 含有不正确内容，请修改后重试(不支持行内注释)`, 'red');
```
ole3021, line 115: 这里有几个问题

* 如果本身的环境变量的值就是带有`#`字符的话，这样处理会产生误会。比如`account=some#sec$re*Id`, 这里 `#`  并不是注释的含义。
* 另外不应该显示用户的环境变量值在log中，因为比如用户配置的是secret，但是因为最最后加入了注释，导致解析失败，会导致这一条log会被系统日志收集，增加了secret暴露的风险。 

建议只修改当前的提示信息清晰提示用户所传的值无效，而不是没有值 就可以了。

ole3021 (changes requested):

src/cli/index.js
ole3021: @zongUMR

zongUMR: 1. I mentioned this concern in my comment above, so how we should check the value here? In the normal `.env` file, syntax will say ` #xxxxx` is a comment(**a space before the #symbol**), should we use the same rule? 
2. Agreed, will update it.

ole3021: >     1. concern

1. The issue I think is `The secretId is not found, please ensure that your secret id is correct.`

This may confuse the user that either the .env file works and why the secretId is not found (actually it is incorrect.)

The root cause of this issue is that the use not understand for dotenv, the comment should start with a new line with #. 

For the solution, I think we can do 

* Fix the error message, make it clear that the value is incorrect instead of key's value not found.  eg:`环境变量${key}的值无效，请核对后重试。`
* Furthermore, if we can identify the value of .env file is incorrect, we can do extra check if there have `#` with several spaces before it and show the warning message eg: `在dotenv配置中发现 #,请确保注释都写在单独由#开头的一行。详情查看：https://github.com/motdotla/dotenv#rules`

zongUMR: 1. Use warning message to alert users
2. Do not exit process
Screen Shot 2021-03-01 at 18 22 34

ole3021 (approved):

medikoo (dismissed):
Looks good to me 👍

PR #908, opened 2021-03-18
Show Chinese --help info for users in China

timqian:
https://app.asana.com/0/1200011502754281/1200052298048317

### DIFF
--- a/src/legacy.js
+++ b/src/legacy.js
@@ -15,7 +15,7 @@ const {
 } = require('./cli/utils');
 
 // These keywords are intercepted by the Serverless Components CLI
-const componentKeywords = new Set(['registry', 'init', 'publish']);
+const componentKeywords = new Set(['registry', 'init', 'publish', '--help', 'help']);
 // These keywords are allowed for nested templates
 const nestedTemplateKeywords = new Set(['deploy', 'remove', 'info']);
 
--- a/src/legacy.js
+++ b/src/legacy.js
@@ -15,7 +15,7 @@ const {
 } = require('./cli/utils');
 
 // These keywords are intercepted by the Serverless Components CLI
-const componentKeywords = new Set(['registry', 'init', 'publish', '--help', 'help']);
+const componentKeywords = new Set(['registry', 'init', 'publish']);
 // These keywords are allowed for nested templates
 const nestedTemplateKeywords = new Set(['deploy', 'remove', 'info']);
 
@@ -35,11 +35,6 @@ const runningComponents = () => {
     return true;
   }
 
-  // Chinese users running "serverless deploy" in a project without a serverless config file
-  if (isChinaUser() && process.argv[2] === 'deploy' && !hasServerlessConfigFile(process.cwd())) {
-    return true;
-  }
-
   try {
     componentConfig = legacyLoadComponentConfig(process.cwd());
   } catch (e) {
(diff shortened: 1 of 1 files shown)


zongUMR (approved):

medikoo (changes requested):
This will break showing Framework help for regular framework users ~~in China~~ (wherever they are). It'll be not accessible to them.

timqian:
@medikoo That's right, thanks for pointing it out, I have updated the PR. Now it only affects users in China.

We decided to show Chinese help command to users in China earlier cc @ole3021 , but I agree it will affect users using v1 version in China (although the interactive init command `sls` already creating a "Component").

So I bring another [PR for serverless](https://github.com/serverless/serverless/pull/9145/files). Which shows a hint about `sls --help-components`. Guess that will be a safer solution.

ole3021:
> regular framework users ~in China~ (wherever they are

Yes, the reason I like to replace Framework help with Components helps for Users from China is because:

*  Most framework commands are not working correctly with Tencent in China.
* The framework will not be supported by Tencent (as infrastructure difference), and Tencent will put more attention on SCF components instead of components v1 feature.
* The framework (plugin feature) is removing from Tencent Serverless Framework docs. 

**To give users a consistent user experience and make it easier for users to start using serverless with Tencent**. I like to

### In non-serverless folder:
> Users can't really do much in non-serverless folders, but get a glance at the commands that to interact with serverless. 
* Use components help command as the default commands help info.
(This will make users keep focusing on components features, also no need for users to understand the difference and switch between framework & components.)

### In components folder:
* Keep showing components help commands. 

### In framework(plugin feature) folder
* Keep using framework commands help info. 
* Add extra help info to let users know that the framework commands are not supported by Tencent,  it works only for users using AWS in China. 

--- 
If we make the changes above, There are the following things that need to be considered:

- [ ] How we like users to create and working with a framework(plugin feature) project on AWS.
  * Optino1:  Using ENV config `SERVERLESS_PLATFORM_VENDOR=aws` to switch the non-serverless folder default help commands.  (with vendor set to AWS, show the framework commands instead. But remind by Tim, the `SERVERLESS_PLATFORM_VENDOR` is designed for components, and it is abuse here in this case. ) 
  * Option2:  Add `create` command to create a framework (plugin feature) project, and interact with framework commands under that folder. (With this, users in China who like to use framework features with AWS still have a way to start.)
  
Please let me know how you think about this, @medikoo  @ac360

medikoo:
@ole3021 I think best approach, is to go with solution proposed here: https://github.com/serverless/serverless/pull/9145

So on `sls --help` in Components v2 folder show Components help. in any other folder show regular Framework help but with first line in chinese, directing users to rely on `sls --help-components` for Components help output.

Thing is that Framework offers many commands dedicated to not work in service context, so it'll be a clear degradation to suddenly not show any Framework related help in such scenario (also afaik Components help does not provide any hint on how to deal with such situation, e.g. how to show Framework related help, as currently is done in Framework help output)

Additionally it'll be good to start thinking on how to totally separate Components CLI from Framework, as both CLI's just harm each other (1) Components CLI, delays start up of Framework CLI by 0.4s (2) Framework CLI confuses Tencent users with non Tencent help output by default.

I believe the only right solution to problems we're facing is to __serve both as two different CLI programs__ (e.g. keep framework as `serverless`, and port Components CLI as e.g. `sc`).

ole3021:
> Framework CLI confuses Tencent users with non Tencent help output

Ok, got that, and let hold on changing about this replacement, and I will plan the seperate solution.  Thanks for the info. 

cc: @timqian

ole3021:
After a discussion with Tencent and Austen.  We confirmed that we won't create a new CLI for Tencent as the serverless brand is quite important and it should not be changed to any other commands. For the traditional CLI experience, Tencent will cooperate with us to make it work for users in China. and we just need to do translation after is ready for users to use. 

will close this PR. 
@medikoo  @timqian

PR #913, opened 2021-03-23
reserve comments for templates

timqian:
> issue: https://app.asana.com/0/1200011502754281/1200078207478206

This PR helps to reserve the comments in the serverless.yml of templates

## Before

`sls init eggjs-starter`
Screen Shot 2021-03-23 at 1 25 58 PM

## After

Screen Shot 2021-03-23 at 1 26 42 PM

### DIFF
--- a/src/cli/commands-cn/init.js
+++ b/src/cli/commands-cn/init.js
@@ -73,7 +73,7 @@ const initTemplateFromCli = async (targetPath, packageName, registryPackage, cli
   const ymlParsed = await parseYaml(serverlessFilePath);
 
   if (appName && ymlParsed.app) {
-    const newYmlContent = ymlOriginal.replace(/^app:\s\S*/m, `app: ${appName}`, 'm');
+    const newYmlContent = ymlOriginal.replace(/^app:\s\S*/m, `app: ${appName}`);
     await fse.writeFile(serverlessFilePath, newYmlContent, 'utf8');
   }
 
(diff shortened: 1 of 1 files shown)


timqian:
The reason we lost the comments of `serverless.yml` after `sls init` a template is because we need to change the `app` key in the yml. Currently, we firstly turn this yml to an `object`, update this object and then recreate a yml. In this process, the comments are lost. 

Unfortunately, `js-yaml` is not able to persist comments while parsing yml: https://github.com/nodeca/js-yaml/issues/549

There are two ways to fix this issue.

### Option 1: 
Use libs that preserve comments.

Disadvatages
- There aren't many libs capable of doing this. After some search we find this lib: [yawn-yaml](https://github.com/mohsen1/yawn-yaml)
- this lib is not very actively maintained, we are using js-yml in other places and replacing this lib may introduce other issues.
- By using the lib above need to add at least 500kb to the components CLI
  - ref: [https://github.com/mohsen1/yawn-yaml/blob/master/package.json](https://github.com/mohsen1/yawn-yaml/blob/master/package.json)
  - [https://www.npmjs.com/package/yawn-yaml](https://www.npmjs.com/package/yawn-yaml)
  - [https://github.com/connec/yaml-js/blob/master/yaml.js](https://github.com/connec/yaml-js/blob/master/yaml.js)
  - [https://www.npmjs.com/package/lodash](https://www.npmjs.com/package/lodash)

### Option 2: 

Use RegExp to replace app(current implementation)

I have tested this on all our templates, it works well.

ole3021 (approved):

PR #916, opened 2021-03-25
fix: runningTemplate logic and check basic validation.

zongUMR:
## What has been implemented?

Closes https://app.asana.com/0/1200011502754281/1200090410758566/f

1. Add more logic for judging a folder is template or not:
    1. All sub folders use same app or stage
    2. All sub folders do not have app or stage
 2. Hoisting some basic validation before user's login in a valid instance project:
     - Check current folder has a yml config.
     - Check name and component fields.

### DIFF
--- a/src/cli/commands-cn/info.js
+++ b/src/cli/commands-cn/info.js
@@ -8,7 +8,7 @@ const path = require('path');
 const { ServerlessSDK } = require('@serverless/platform-client-china');
 const { v4: uuidv4 } = require('uuid');
 const utils = require('./utils');
-const { runningTemplate } = require('../utils');
+const { runningTemplate, checkTemplateAppAndStage } = require('../utils');
 const infoAll = require('./infoAll');
 const chalk = require('chalk');
 const moment = require('moment');
@@ -17,11 +17,11 @@ module.exports = async (config, cli, command) => {
   // Start CLI persistance status
   cli.sessionStart('Initializing', { timer: false });
 
-  if (runningTemplate(process.cwd(), true)) {
+  let instanceDir = process.cwd();
+  if (runningTemplate(instanceDir) && checkTemplateAppAndStage(instanceDir)) {
     return infoAll(config, cli);
   }
 
-  let instanceDir = process.cwd();
   if (config.target) {
     instanceDir = path.join(instanceDir, config.target);
   }
--- a/src/cli/commands-cn/info.js
+++ b/src/cli/commands-cn/info.js
@@ -17,17 +17,18 @@ module.exports = async (config, cli, command) => {
   // Start CLI persistance status
   cli.sessionStart('Initializing', { timer: false });
 
-  await utils.login();
-
-  if (runningTemplate(process.cwd())) {
+  if (runningTemplate(process.cwd(), true)) {
     return infoAll(config, cli);
   }
 
-  // Load YAML
   let instanceDir = process.cwd();
   if (config.target) {
     instanceDir = path.join(instanceDir, config.target);
   }
+  await utils.checkBasicConfigValidation(instanceDir);
+  await utils.login();
+  // Load YAML
+
   const instanceYaml = await utils.loadInstanceConfig(instanceDir, command);
 
   // Presentation
--- a/src/cli/commands-cn/run.js
+++ b/src/cli/commands-cn/run.js
@@ -6,7 +6,12 @@
 
 const path = require('path');
 const fs = require('fs');
-const { runningTemplate, loadInstanceConfig, fileExists } = require('../utils');
+const {
+  runningTemplate,
+  loadInstanceConfig,
+  fileExists,
+  checkTemplateAppAndStage,
+} = require('../utils');
 const { ServerlessSDK, utils: tencentUtils } = require('@serverless/platform-client-china');
 const { v4: uuidv4 } = require('uuid');
 const utils = require('./utils');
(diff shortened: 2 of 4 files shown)


src/cli/utils.js
```diff
@@ -521,6 +538,20 @@ const runningTemplate = (root) => {
         }
       }
     }
+
+    // If we need to check the app and stage value for all sub folders, if we have more than 1 app value or at least one folder's yml has app but others do not have, we think it's not a valid template project
```
ole3021, line 542: This is incorrect. The template subfolder will inherit the `app` and `stage` from the main folder. 

* folder `M` has subFolders `A`, `B`

is a template folder cases:

* `M`, has a yaml config with app and name. 
  * app and name in `A` and `B` has no difference. eg: A has app and stage(same to the value in M), B has no app and stage.
    * `A` and `B` has no app and stage
    * `A` and `B` have the same app and stage.
* `M`, has no yaml config 
  * `A` and `B` should all has app and stage with the same value.

otherwise, it is not a valid template folder.  include cases like

* `M` has no config, `A` and  `B` has different app name or stage.

src/cli/utils.js
```diff
@@ -497,7 +497,7 @@ const isDeployableProjectExists = (inputPath) => {
   return false;
 };
 
-const runningTemplate = (root) => {
+const runningTemplate = (root, checkAppAndStage) => {
```
ole3021, line 500: the `checkAppAndStage` field make no sense in my mind. 

* this fun is to check if the root is a template, and we just like add extra validations to make it more correct.  Is there a case that a folder does not match the `appAndStage` rule but still a valid template folder?
* The checkAppAndStage we don't know what's the type and how it works for running template directly. add a default value to show is a boolean if necessary.

src/cli/commands-cn/info.js
```diff
@@ -17,17 +17,18 @@ module.exports = async (config, cli, command) => {
   // Start CLI persistance status
   cli.sessionStart('Initializing', { timer: false });
 
-  await utils.login();
-
-  if (runningTemplate(process.cwd())) {
+  if (runningTemplate(process.cwd(), true)) {
     return infoAll(config, cli);
```
ole3021, line 21: Can we confirm there is not user info required in the method `infoAll`, as previously we have this issue when moving login after the logic. 

Also, I think we can simply check `isInvalidPath(root)` and throw the invalid error if is not a serverless project or template folder. With this, we can just add an extra check without changing our existing logic.  Current issue is that

* A folder contains multiple serverless projects (obviously not a template as well) considered as a template folder during the run time.

src/cli/commands-cn/run.js
```diff
@@ -40,13 +40,16 @@ module.exports = async (config, cli, command) => {
...
-
-  // Load YAML
   let instanceDir = process.cwd();
   if (config.target) {
     instanceDir = path.join(instanceDir, config.target);
   }
+
+  await utils.checkBasicConfigValidation(instanceDir);
+
+  await utils.login();
```
ole3021, line 50: Same above, can we confirm that the logic in `runAll` not required the login info exist? @timqian  please help double check.

src/cli/commands-cn/utils.js
```diff
@@ -43,6 +43,26 @@ const getDefaultOrgName = async () => {
...
 
+/**
+ * check basic config validation
+ */
+
+const checkBasicConfigValidation = async (dicPath) => {
+  const instanceFile = loadInstanceConfig(dicPath);
+
+  if (!instanceFile) {
+    throw new Error('serverless config file was not found');
```
ole3021, line 54: `没有找到serverless配置文件，请检查。`

src/cli/commands-cn/utils.js
```diff
@@ -43,6 +43,26 @@ const getDefaultOrgName = async () => {
...
+  }
+
+  if (!instanceFile.name) {
+    throw new Error('Missing "name" property in serverless.yml');
```
ole3021, line 58: `在serverless配置文件中没有发现实例名称("name"字段)，请检查。`

src/cli/commands-cn/utils.js
```diff
@@ -43,6 +43,26 @@ const getDefaultOrgName = async () => {
...
+  }
+
+  if (!instanceFile.component) {
+    throw new Error('Missing "component" property in serverless.yml');
```
ole3021, line 62: `在serverless配置文件中没有发现组件类型("component"字段)，请检查。
`

ole3021 (dismissed):

zongUMR:
@timqian This RP add some Chinese translations

src/cli/commands-cn/utils.js
```diff
@@ -51,15 +71,15 @@ const loadTencentInstanceConfig = async (directoryPath, command) => {
...
   }
 
   if (!instanceFile.name) {
-    throw new Error('Missing "name" property in serverless.yml');
+    throw new Error('在serverless配置文件中没有发现实例名称("name"字段)，请检查。');
   }
 
   if (!instanceFile.component) {
-    throw new Error('Missing "component" property in serverless.yml');
+    throw new Error('在serverless配置文件中没有发现组件类型("component"字段)，请检查。');
```
timqian, line 82: Looks we can reuse the `checkBasicConfigValidation` function here

src/cli/utils.js
```diff
@@ -497,6 +497,52 @@ const isDeployableProjectExists = (inputPath) => {
...
+            apps.add(instanceYml.app);
+          } else {
+            hasNullApp = true;
+          }
+        }
+      }
+    }
+
+    // If we need to check the app and stage value for all sub folders, if we have more than 1 app value or at least one folder's yml has app but others do not have, we think it's not a valid template project
+    // same to stage value. Valid template standard: 1. All sub folders do not have app and stage; 2. All sub folders have one same app and stage.
```
timqian, line 534: Will this limitation make some users' current nested project undeployable?

timqian:
@zongUMR It will be great if you can update the corresponding integration tests here https://github.com/serverlessinc/integration-test-cn

src/cli/utils.js
zongUMR: I do not think so. this function only hoist some validation before login. If user's project is deployable (template or instance),  it will not block anything.

timqian (approved):

PR #917, opened 2021-03-25
Check latest components version and ask users to upgrade

timqian:
> https://app.asana.com/0/1200011502754281/1200090410758582

## What has been implemented?

Check the latest Components version and recommend users to upgrade if they are using an older version.
It works similar to npm, when a user is using an older version of npm, npm will recommend the user to upgrade the CLI:

Screen Shot 2021-03-25 at 11 38 19 AM

I am aware that for the traditional CLI, we have introduced an `autoUpdate.enabled` configuration for users, users can use this setting to decide if they would like the CLI to check the latest version and upgrade itself after executing commands. ([PR](https://github.com/serverless/serverless/pull/8428)). 

But as we don't have a global configuration file for Components to store user's settings, I think recommending users to upgrade the CLI themselves might be a good idea

## Other info
I used the [latest-version](https://github.com/sindresorhus/latest-version) package to check latest version of components, because directly requesting `registry.npmjs.org` may be slow for many Chinese users, and [latest-version](https://github.com/sindresorhus/latest-version) is able to request to the registry users configured for themselves.

### DIFF
--- a/src/cli/index.js
+++ b/src/cli/index.js
@@ -12,9 +12,9 @@ const dotenv = require('dotenv');
 const semver = require('semver');
 const chalk = require('chalk');
 const HttpsProxyAgent = require('https-proxy-agent');
-const getLatestVersion = require('latest-version');
 const CLI = require('./CLI');
 const { loadInstanceConfig, fileExistsSync, isProjectPath, isChinaUser } = require('./utils');
+const updateNotifier = require('update-notifier');
 
 module.exports = async () => {
   const args = minimist(process.argv.slice(2));
@@ -145,6 +145,17 @@ module.exports = async () => {
     commands = require('./commands-cn');
     InvalidCommandMsg =
       "检测到当前目录下已有 serverless 项目，请通过 'sls deploy' 进行部署，或在新路径下完成 serverless 项目初始化";
+
+    // For non China versions we have upgrade notifications configured through backend notifications service
+    const notifier = updateNotifier({
+      pkg: require('../../package.json'),
+      // Show update notification one time a day at most, this is done to not be annoying to the user
+      updateCheckInterval: 1000 * 60 * 60 * 24,
+    });
+
+    notifier.notify({
+      message: '发现 serverless 新版本，建议通过 `npm update -g serverless` 进行升级',
+    });
   } else {
     commands = require('./commands');
   }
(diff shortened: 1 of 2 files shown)


zongUMR:
@timqian I agree that we need to only notify user to update CLI by `npm -g i serverless`, just like the npm, and here is another package: [update-notifer](https://www.npmjs.com/package/update-notifier) which can output the info directly.

timqian:
@zongUMR Thanks for the note! directly using `update-notifer` looks like a better option

medikoo (changes requested):
For non China versions we already have upgrade notifications configured through backend notifications service. Technically same could be done in context of `platform-cn`.
Anyway I think it's ok to do it here, but only if it's applied to users in China (while as I see this functionality is run unconditionally for any users).

Also `npm install -g severless` is not guranteed to upgrade `@serverless/components` which is a dependency of `serverless`. Bulletproof would be `npm update -g serverless`.

Additionally notifier should only notify of minor upgrades. Note that `major` upgrades will in all cases be impossible that way (as `@serverless/components`  version in `serverless` is always pinned to specific major). In our backend notification service we have intelligence which ensures that user is notified about updates in major they're relying on(e.g. our current latest is v2, but we created one v1 release, and v1 users were notified about it, even though it's not _latest_ on npm grounds)

timqian:
@zongUMR @medikoo I have updated the PR a bit by using `[update-notifer](https://github.com/yeoman/update-notifier#how)`:

- [x] Whenever a Chinese user uses the CLI, it asynchronously checks with npm in the background for available updates and won't block 
- [x] Notification is shown to users one time a day at most, this is done to not be annoying to the user
- [x] Notification will be shown when there is `major`, `minor`, or `patch` upgrades
- [x] Notification preview

Screen Shot 2021-03-29 at 11 14 36 AM

src/cli/index.js
```diff
@@ -14,6 +14,7 @@ const chalk = require('chalk');
 const HttpsProxyAgent = require('https-proxy-agent');
 const CLI = require('./CLI');
 const { loadInstanceConfig, fileExistsSync, isProjectPath, isChinaUser } = require('./utils');
+const updateNotifier = require('update-notifier');
```
medikoo, line 17: Let's require it only when it's used, so let's put it into that `if block

medikoo (commented):
@timqian looks good, I've proposed to addd just one minor optimization for non-china users

medikoo (approved):
Thanks @timqian looks good

zongUMR (approved):

ole3021:
Sorry for responding late,  As we just researched the notification features, and if we don't implement the update automatically in the background.  I think we need to consider the two options below. 

Option A: Implement in this PR. 
* Implement on the CLI side, and show a message for users to update to the latest version. 

Option B: Use the notificaiton.
* Implement on the Server-side (sls-metrics), and show a message only after use deploy and under a specific version. 

I prefer to go with the Option B:
* The CLI implementation is not working for older versions, but the notification will. 
* we don't need to ask user to update for each patch version release.
* we better check the version for `serverless/cli` and `serverless/components` separately. 

@timqian  @zongUMR  @medikoo  ^

medikoo:
> I prefer to go with the Option B:

I agree, it gives better control and also works for older version.

You may take an inspiration from `@serverlessinc/platform` repo, on how this notification is resolved there (If you lack access I'll be happy to expose needed files)

timqian:
I agree, notifying users using older versions is definitely better. I will take a look at the `platform` code and try to add it to `platform-cn`

ole3021:
Close the CLI change for update warning, will use the notification feature to show user update warning messages.

PR #924, opened 2021-04-01
feat: add notification and update deps

zongUMR:
## What has been implemented?

Closes https://app.asana.com/0/1200011502754281/1200140904325774/f

1. Update `serverless/utils` to 4.x
2. Add **questionnaire** notification for the above task

## Steps to verify

- Run X
- ...

## Todos:

- [ ] Write tests
- [ ] Run Prettier

### DIFF
--- a/src/cli/commands-cn/run.js
+++ b/src/cli/commands-cn/run.js
@@ -107,10 +107,7 @@ module.exports = async (config, cli, command) => {
   let deferredNotificationsData;
   if (command === 'deploy') {
     deferredNotificationsData = requestNotification(
-      Object.assign(generateNotificationsPayload(instanceYaml), {
-        command: 'deploy',
-        actions: [utils.notificationActions.components_questionnaire],
-      })
+      Object.assign(generateNotificationsPayload(instanceYaml), { command: 'deploy' })
     );
 
     // Warn about dev agent
--- a/src/cli/commands-cn/run.js
+++ b/src/cli/commands-cn/run.js
@@ -107,7 +107,10 @@ module.exports = async (config, cli, command) => {
   let deferredNotificationsData;
   if (command === 'deploy') {
     deferredNotificationsData = requestNotification(
-      Object.assign(generateNotificationsPayload(instanceYaml), { command: 'deploy' })
+      Object.assign(generateNotificationsPayload(instanceYaml), {
+        command: 'deploy',
+        actions: [utils.notificationActions.components_questionnaire],
+      })
     );
 
     // Warn about dev agent
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
   "dependencies": {
     "@serverless/platform-client": "^4.2.2",
     "@serverless/platform-client-china": "^2.1.9",
-    "@serverless/utils": "^3.1.0",
+    "@serverless/utils": "^4.0.0",
     "adm-zip": "^0.5.4",
     "ansi-escapes": "^4.3.1",
     "chalk": "^4.1.0",
--- a/src/cli/commands-cn/utils.js
+++ b/src/cli/commands-cn/utils.js
@@ -446,6 +446,10 @@ inputs:
   throw new Error('当前目录未检测到 Serverless 配置文件');
 };
 
+const notificationActions = {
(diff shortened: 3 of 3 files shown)


src/cli/commands-cn/run.js
```diff
@@ -107,7 +107,10 @@ module.exports = async (config, cli, command) => {
   let deferredNotificationsData;
   if (command === 'deploy') {
     deferredNotificationsData = requestNotification(
-      Object.assign(generateNotificationsPayload(instanceYaml), { command: 'deploy' })
+      Object.assign(generateNotificationsPayload(instanceYaml), {
+        command: 'deploy',
+        actions: [utils.notificationActions.components_questionnaire],
```
medikoo, line 112: This looks as a static property, why exactly it's added?

medikoo (commented):
Thanks @zongUMR ! I have just one question and we should be good

src/cli/commands-cn/run.js
zongUMR: The backend service will use actions to insert different notify messages, you can the the backedn PR here: https://github.com/serverlessinc/platform-cn/pull/165/files

medikoo (commented):
> The backend service will use actions to insert different notify messages, you can see the the backend PR here: https://github.com/serverlessinc/platform-cn/pull/165/files

Thanks for explanations, still it looks as if it try to go against the design of backend notification service.

The way it is designed, is that depending on the payload data (which describes deployed service characteristics, and provides basic installation info), it's _backend_ service that decides which notification to show. e.g. if backend will notice that payload reflects outdated version, it returns version update notification.

While what you proposed here is that client instead of sending service characteristics in a payload, asks for specific notification. Is my understanding correct?

What's the point of sending such information from client?

zongUMR:
> > The backend service will use actions to insert different notify messages, you can see the the backend PR here: https://github.com/serverlessinc/platform-cn/pull/165/files
> 
> Thanks for explanations, still it looks as if it try to go against the design of backend notification service.
> 
> The way it is designed, is that depending on the payload data (which describes deployed service characteristics, and provides basic installation info), it's _backend_ service that decides which notification to show. e.g. if backend will notice that payload reflects outdated version, it returns version update notification.
> 
> While what you proposed here is that client instead of sending service characteristics in a payload, asks for specific notification. Is my understanding correct?
> 
> What's the point of sending such information from client?

Got it, this is only a static notify, it does not need to send anything to backend. I need to only show the notification in CLI. A new commit fix it. Thanks @medikoo

src/cli/commands-cn/run.js
```diff
@@ -164,7 +167,15 @@ module.exports = async (config, cli, command) => {
   }
   cli.sessionStop('success', 'Success');
 
-  if (deferredNotificationsData) printNotification(cli, await deferredNotificationsData);
+  if (deferredNotificationsData) {
```
timqian, line 170: Looks like this static notification will only be shown when we got a backend notification. Is this what we want?

ole3021:
@zongUMR  We should not implement sending the questionnaire logic on the client (CLI) side. 

if I'm understanding right, we have to trigger the server-side notification with a `NotificationsPayload`, but I'm not sure what's really the static notification is hrere. 

The ideal notification process is that

After the user make a deployment, and the server can send back a message to the user CLI console with the notification message that we like to show to the user. Also 
* This should work with back-compatible, no matter which version the user is using should get the messages.
* We can controller easier to add, change, remove the messages, without change anything on the CLI side.
    * so that we can add a new message, and remove messages as we want. CLI doesn't need to know about that.

Also from our design principle: We should keep our CLI simple clean and not include any business logic, all the logic should be not related to a specific version of CLI but implement on the backend side as much as possible.

src/cli/commands-cn/run.js
zongUMR: This is the current behavior, we will not change it.

zongUMR:
@medikoo @ole3021 @timqian This PR will not change any code, only update a package which includes the **China dev notify url** for testing: https://github.com/serverless/utils/pull/78, the real logic will put on server side

timqian (approved):

PR #936, opened 2021-04-22
Update logo

timqian:
Issue: https://app.asana.com/0/1200011502754281/1200199437834091
@ole3021 suggests that we should update the logo in the components. But this will also affect global users. What do you think @ac360 @eahefnawy @medikoo

### DIFF
--- a/src/cli/CLI.js
+++ b/src/cli/CLI.js
@@ -376,8 +376,8 @@ class CLI {
   logLogo() {
     let logo = os.EOL;
     logo += 'serverless';
-    logo += red(' ⚡');
-    logo += 'framework';
+    logo += red(' ⚡ ');
+    logo += 'components';
 
     if (process.env.SERVERLESS_PLATFORM_STAGE === 'dev') {
       logo += grey(' (dev)');
@@ -392,7 +392,7 @@ class CLI {
   logRegistryLogo(text) {
     let logo = os.EOL;
     logo += white('serverless');
-    logo += red(' ⚡');
+    logo += red(' ⚡ ');
     logo += white('registry');
 
     if (process.env.SERVERLESS_PLATFORM_STAGE === 'dev') {
--- a/src/cli/CLI.js
+++ b/src/cli/CLI.js
@@ -376,7 +376,7 @@ class CLI {
   logLogo() {
     let logo = os.EOL;
     logo += 'serverless';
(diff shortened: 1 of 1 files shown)


src/cli/CLI.js
```diff
@@ -376,8 +376,8 @@ class CLI {
   logLogo() {
     let logo = os.EOL;
     logo += 'serverless';
-    logo += red(' ⚡');
-    logo += 'framework';
+    logo += red(' ⚡ ');
+    logo += 'components';
```
medikoo, line 380: @timqian that looks good to me. Still I believe @ac360 and @eahefnawy have to approve that.

I see that you've also added whitespace, can you post a screenshot showing difference introduced?

timqian: Before
Screen Shot 2021-04-22 at 3 39 02 PM

After
Screen Shot 2021-04-22 at 3 38 41 PM

timqian: But this does not look good in vscode console. I guess we should not add a space @ole3021 

Screen Shot 2021-04-22 at 3 42 04 PM

medikoo: Yes, it's probably better to keep it the way it was

ole3021: @medikoo  It does looks without issue in VS code without the extra space. And I'm ok if we like to keep it the way it was.(i use terminal than the build-in terminal in VS code)

For the other change from `framework` to `components` the purpose is: I like to notify that users are using the components commands instead of framework. as 
  * compoents commands are different from the framework(core CLI) commands
  * For the documents updating now, we like to tell user the difference about components when user CLI, but there are no way to let user aware about that now.

ole3021: Update the title show in Hyper.term and Terminal of macOS

image

image

medikoo: > I like to notify that users are using the components commands instead of framework

@ole3021 I like that as well. It was just white-space change that I thought it's better to keep as it was.

ole3021: oky, then let's remove the space in the change and just change the name will be fine. @timqian 

thx for the help.

timqian: Just removed the space

zongUMR (approved):

ole3021 (approved):

PR #951, opened 2021-05-18
Replace moment with dayjs

timqian:
As momentjs is not been actively maintained and it has a big size, we replaced it with dayjs in this PR.

A more direct reason we need to replace it is that momentjs prints unnecessary error logs when it is initialized with a wrong formatted date string. By using dayjs we can avoid this.

Screen Shot 2021-05-18 at 11 45 40 AM


medikoo (approved):
Looks good to me.

@timqian were changes tested manually for all areas? Unfortunately this project has not tests setup, so we can't confirm through automated testing that no accidental regression was introduced

timqian:
@medikoo Yes, I have tested `sls info` for normal and nested components projects

PR #961, opened 2021-06-01
feat: invoke local for scf python

zongUMR:
## What has been implemented?

Closes https://app.asana.com/0/1200011502754281/1200376732735304/f

## Steps to verify

- Run X
- ...

## Todos:

- [ ] Write tests
- [x] Run Prettier

### DIFF
--- a/src/cli/commands-cn/invoke/invoke-local/runPython.js
+++ b/src/cli/commands-cn/invoke/invoke-local/runPython.js
@@ -32,9 +32,11 @@ f.close()`;
       const errData = data.toString().split('\n');
       errData.splice(1, 2);
 
+      cli.log('---------------------------------------------');
       fse.unlinkSync(tempPyFile);
       fse.unlinkSync(tempResFile);
-      colorLog(`Python 函数运行错误：${errData.join('\n').toString()}`, 'yellow', cli);
+      colorLog(`调用错误\n\n ${errData.join('\n').toString()}`, 'red', cli);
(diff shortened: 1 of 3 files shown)


src/cli/commands-cn/invoke/invoke-local/runPython.js
```diff
@@ -0,0 +1,52 @@
...
+
+const { spawn } = require('child_process');
+const path = require('path');
+const { v4 } = require('uuid');
+const fse = require('fs-extra');
+const { printOutput, colorLog } = require('./utils');
+
+module.exports = async (event, context, handlerFile, handlerFunc, cli) => {
+  const tempResFile = `serverless_res_${v4().split('-')[0]}.txt`;
+  const tempPyFile = `serverless_tmep_${v4().split('-')[0]}.py`;
```
timqian, line 11: Curious, why do we need to create temp files to execute users' code?

zongUMR: 1. A normal python file in https://github.com/serverless-components/tencent-examples/blob/master/scf-python/src/index.py, it only contains the definition of function, so we must call it somewhere and the **tempPyFile** will import it and run it, store result finally.
2. We need to store the returned result from the called function, we can not directly get the result from a python function calling in the nodejs's child_process, this's why I need **tempResFile** to store the function's result.

timqian (approved):

PR #982, opened 2021-07-01
Refactoring sls logs & invoke

timqian:
Related tickets: 
- https://app.asana.com/0/1200011502754281/1200495858241492
- https://app.asana.com/0/1200011502754281/1200451265990936
- https://app.asana.com/0/1200011502754281/1200520309783654

### DIFF
--- a/src/cli/commands-cn/invoke/index.js
+++ b/src/cli/commands-cn/invoke/index.js
@@ -140,11 +140,6 @@ module.exports = async (config, cli, command) => {
         const retJson = JSON.parse(retMsg);
         cli.log(inspect(retJson, { depth: Infinity, colors: true, compact: 0 }));
       } catch (error) {
-        await storeLocally({
-          ...telemtryData,
-          outcome: 'failure',
-          failure_reason: error.message,
-        });
         cli.log(retMsg);
       }
     } else {
--- a/src/cli/commands-cn/invoke/index.js
+++ b/src/cli/commands-cn/invoke/index.js
@@ -27,18 +27,7 @@ module.exports = async (config, cli, command) => {
     return invokeLocal(config, cli, command);
   }
 
-  const {
-    stage,
-    s,
-    region,
-    r,
-    data,
-    d,
-    path,
-    p,
-    function: originalFunctionAlias,
-    f,
-  } = config;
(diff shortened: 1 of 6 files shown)


src/cli/commands-cn/invoke/index.js
```diff
@@ -136,11 +125,6 @@ module.exports = async (config, cli, command) => {
         const retJson = JSON.parse(retMsg);
         cli.log(inspect(retJson, { depth: Infinity, colors: true, compact: 0 }));
       } catch (error) {
-        await storeLocally({
```
zongUMR, line 139: Why remove `storeLocally` here?

timqian: Because this is not a `failure`, sometimes the retMsg is not a JSON and we want to display it directly

src/cli/index.js
```diff
@@ -198,8 +198,9 @@ module.exports = async () => {
     }
   } catch (error) {
     process.exitCode = 1;
-
-    if (cli.isSessionActive()) {
+    if (error.isWarning) {
+      cli.logWarning(error);
```
timqian, line 202: @medikoo We would like to display some types of errors differently, so we added a new flag in the error sent from engine and added this check here. cc @ole3021

src/cli/commands-cn/invoke/index.js
zongUMR: Got it, but plz do not remove it, just remove the **outcome** and **failure_reason** fields, keep the record saving

timqian: @zongUMR This block of code is just for checking if the `retMsg` is a JSON or not, why do you think we need to save the record here? (BTW, we have the record saving logic here https://github.com/serverless/components/pull/982/files#diff-2a4f4f9def1700b7cc7cf1d95bf2250063ea626cf3bc4a69101cc51cefbe796fR134)

zongUMR: Ok , just ensure we can save the data locally, no matter success or fail

medikoo (approved):
Looks good to me 👍

zongUMR (approved):

PR #990, opened 2021-07-13
fix: hacker way to solve cn dev command doesn't work sometime

bodyno:
https://app.asana.com/0/1200011502754281/1200168199767455/f

Hacker way to solve cn dev command doesn't work sometime.

Use setTimeout to solve this problem now. Later will be careful find a better way to solve it.

image


zongUMR:
Could you explain why the **dev mode** does not work sometimes and we can use **setTimeout** to resolve it?

bodyno:
> Could you explain why the **dev mode** does not work sometimes and we can use **setTimeout** to resolve it?

Now use setTimeout can solve problem. The bug is on server side. I will keep find and solve the problem on server side.

PR #992, opened 2021-07-14
Ignore timeout issue when polling logs

timqian:
- [x] Ignore errors when polling logs with `sls logs -t`
- [x] Remove unnecessary code blocks

Related ticket
- https://app.asana.com/0/1200011502754281/1200601499823399

### DIFF
--- a/src/cli/commands-cn/logs.js
+++ b/src/cli/commands-cn/logs.js
@@ -116,26 +116,10 @@ module.exports = async (config, cli, command) => {
       return logs;
     } catch (error) {
       telemtryData.outcome = 'failure';
-
-      if (error.code === '1001') {
-        telemtryData.failure_reason =
-          '无法找到指定 SCF 实例，请检查 SCF 实例名称和 Stage / Region 信息或重新部署后调用';
-
-        cli.log(
-          `Serverless: ${chalk.yellow(
-            '无法找到指定 SCF 实例，请检查 SCF 实例名称和 Stage / Region 信息或重新部署后调用'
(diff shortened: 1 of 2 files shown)


zongUMR (approved):

src/cli/commands-cn/logs.js
```diff
@@ -156,7 +140,13 @@ module.exports = async (config, cli, command) => {
 
     cli.sessionStart('监听中');
     setInterval(async () => {
-      const newLogList = await getLogList();
+      let newLogList;
+      // ignore timeout issue when polling new logs
+      try {
+        newLogList = await getLogList();
+      } catch (error) {
```
zongUMR, line 147: Can't we fix this error, or just ignore it?

timqian: As you may already know, there is a small chance that users will meet timeout issues when requesting backend APIs due to network issues. We cannot do much on it. And for `sls logs -t`, by default, we will poll `getLogs` API every 2 seconds, if some of the requests timeout, we should not show the unhandled promise rejection in users console. Ignoring them should be the best way.

zongUMR: A question: If there is a non-timeout error occur here, but we just `return 0`, and can not catch those kinds of error anymore?

timqian: When users execute `sls logs -t`, we will do a `getLogList` request first and if there are any "non-timeout" issues, we should have been able to catch the issue.

PR #993, opened 2021-07-15
Add --target support for 3 commands in template dir

timqian:
- [x] `--target` support for for `invoke`, `invoke local` and `logs` command
- [x] `--target` in help doc for the above commands
- [x] Notify users to use `--target` option if they are using these commands directly in a composite project 

### Related tickets
https://app.asana.com/0/1200011502754281/1200578828969029


zongUMR (approved):
LGTM, but I think we do not need `checkTemplateAppAndStage` here, we do not need to if it's a valid template project, just need to confirm it's a template project

timqian:
@zongUMR Thanks, I agree that `checkTemplateAppAndStage` is not necessary, just removed it

zongUMR (approved):

PR #994, opened 2021-07-15
Improve component logic

zongUMR:
Close: https://app.asana.com/0/1200011502754281/1200595402967504/f
---
1. [x] If Chinese user is in component or template project, we just use components logic. @medikoo  I think the change in `legacy.js` will not affect global code, but please take a loot
2. [x] Add **--target** option for `dev` command
3. [x] Check **path** validation

### DIFF
--- a/src/legacy.js
+++ b/src/legacy.js
@@ -25,6 +25,11 @@ const runningComponents = () => {
   let componentConfig;
   let instanceConfig;
 
+  // For Chinese users, if they are run a command inside component or template project, just use components to execute it
+  if (isChinaUser() && (hasServerlessConfigFile(process.cwd()) || runningTemplate(process.cwd()))) {
+    return true;
+  }
+
   // load components if user runs a keyword command, or "sls --all" or "sls --target" (that last one for china)
   if (
     componentKeywords.has(process.argv[2]) ||
--- a/src/legacy.js
+++ b/src/legacy.js
@@ -17,18 +17,22 @@ const {
 // These keywords are intercepted by the Serverless Components CLI
 const componentKeywords = new Set(['registry', 'init', 'publish']);
 // These keywords are allowed for nested templates
-const nestedTemplateKeywords = new Set(['deploy', 'remove', 'info', 'help', '--help']);
+const nestedTemplateKeywords = new Set([
+  'deploy',
+  'remove',
+  'info',
+  'help',
+  '--help',
+  'dev',
+  'logs',
+  'invoke',
+]);
 
 const runningComponents = () => {
   const args = minimist(process.argv.slice(2));
 
   let componentConfig;
   let instanceConfig;
-
-  // For Chinese users, if they are run a command inside component or template project, just use components to execute it
-  if (isChinaUser() && (hasServerlessConfigFile(process.cwd()) || runningTemplate(process.cwd()))) {
-    return true;
-  }
 
   // load components if user runs a keyword command, or "sls --all" or "sls --target" (that last one for china)
   if (
--- a/src/cli/commands-cn/dev.js
+++ b/src/cli/commands-cn/dev.js
@@ -12,6 +12,8 @@ const { ServerlessSDK, utils: chinaUtils } = require('@serverless/platform-clien
 const { generatePayload, storeLocally } = require('./telemtry');
 const { v4: uuidv4 } = require('uuid');
 const utils = require('./utils');
+const chalk = require('chalk');
+const { runningTemplate } = require('../utils');
 
 class LogForwardingOutput extends Writable {
   _write(chunk, encoding, callback) {
@@ -138,6 +140,12 @@ module.exports = async (config, cli, command) => {
   let instanceDir = process.cwd();
   if (config.target) {
     instanceDir = path.join(instanceDir, config.target);
+  }
+  if (runningTemplate(instanceDir)) {
+    cli.log(
+      `Serverless: ${chalk.yellow('该命令暂不支持对多组件进行调用, 使用 --target 指定执行目录')}`
+    );
+    process.exit();
   }
 
   const projectFile = await utils.checkBasicConfigValidation(instanceDir);
--- a/src/cli/commands-cn/help.js
+++ b/src/cli/commands-cn/help.js
@@ -94,6 +94,7 @@ ${description(`    --stage / -s             指定环境名称，默认使用配
 ${command2('dev')}                       启动调试模式
 ${description(`    --stage / -s             指定环境名称，默认使用配置环境
     --profile                使用指定身份的全局授权信息
+    --target                 指定执行命令的组件实例路径
 `)}`,
     'logs': `
 ${command2('logs')}                      查看应用日志
--- a/src/cli/commands-cn/utils.js
+++ b/src/cli/commands-cn/utils.js
@@ -59,6 +59,10 @@ const getDefaultOrgName = async () => {
 const checkBasicConfigValidation = async (dicPath) => {
   const instanceFile = loadInstanceConfig(dicPath);
 
+  if (!fse.existsSync(dicPath)) {
+    throw new Error(`指定的路径 ${dicPath} 不存在，请检查后重试`);
+  }
+
   if (!instanceFile) {
     throw new Error('没有找到serverless配置文件，请检查。');
   }


src/legacy.js
```diff
@@ -25,6 +25,11 @@ const runningComponents = () => {
   let componentConfig;
   let instanceConfig;
 
+  // For Chinese users, if they are run a command inside component or template project, just use components to execute it
+  if (isChinaUser() && (hasServerlessConfigFile(process.cwd()) || runningTemplate(process.cwd()))) {
```
medikoo, line 29: It'll be nice if it's combined with conditional which now is at L45. So code goes as:

```javascript
if (isChinaUser()) {
 // China specific conditions
}
```

medikoo (changes requested):
@zongUMR thanks for update. I've proposed some logic organization improvement.

Also it's not clear to me what use case this addresses (?)

In which case we currently do not load components, and after this change we will load them?

src/legacy.js
zongUMR: I will update it

zongUMR:
> @zongUMR thanks for update. I've proposed some logic organization improvement.
> 
> Also it's not clear to me what use case this addresses (?)
> 
> In which case we currently do not load components, and after this change we will load them?

Currently If we use `sls dev` inside a template project(https://github.com/serverless-components/tencent-examples/tree/master/fullstack), it will not load components

medikoo:
> Currently If we use sls dev inside a template project(https://github.com/serverless-components/tencent-examples/tree/master/fullstack), it will not load components

Is this specific just to `sls dev` command? If it's the case, maybe we could narrow condition so it applies just to that (?)

zongUMR:
> > Currently If we use sls dev inside a template project(https://github.com/serverless-components/tencent-examples/tree/master/fullstack), it will not load components
> 
> Is this specific just to `sls dev` command? If it's the case, maybe we could narrow condition so it applies just to that (?)

Actually not, we also have some other commands which should load components: `invoke`, `invoke local`, `logs` and future commands. This Tim's PR: https://github.com/serverless/components/pull/993, it will depend on this change as well

src/legacy.js
zongUMR: I combined those conditions for Chinese users

src/legacy.js
```diff
@@ -52,10 +47,23 @@ const runningComponents = () => {
     // ignore
   }
 
-  if (!componentConfig && !instanceConfig) {
+  // For Chinese users
```
medikoo, line 50: This comment feels obsolete as code is self explanatory on that

src/legacy.js
```diff
@@ -52,10 +47,23 @@ const runningComponents = () => {
...
+  if (isChinaUser()) {
+    // Chinese users running "serverless deploy" in a project without a serverless config file
```
medikoo, line 52: I would also remove this comment, as it code feels self explanatory

src/legacy.js
```diff
@@ -52,10 +47,23 @@ const runningComponents = () => {
...
+
     // When no in service context and plain `serverless` command, return true when user in China
     // It's to enable interactive CLI components onboarding for Chinese users
-    return process.argv.length === 2 && isChinaUser();
+    if (!componentConfig && !instanceConfig && process.argv.length === 2) {
+      return true;
+    }
+
+    // if they are run a command inside component or template project, just use components to execute it
+    if (hasServerlessConfigFile(process.cwd()) || runningTemplate(process.cwd())) {
```
medikoo, line 64: I believe it should be just `if (runningTemplate(process.cwd()))`, otherwise we're introducing a breaking change for all AWS users in China, as it'll unconditionally run Components in context of Serverless Framework services

medikoo (changes requested):
Thanks @zongUMR I see just one issue, otherwise it looks good

src/legacy.js
zongUMR: Fixed.

src/legacy.js
zongUMR: Fixed.

src/legacy.js
zongUMR: Agree, updated it

src/legacy.js
```diff
@@ -52,10 +47,18 @@ const runningComponents = () => {
...
+  if (isChinaUser()) {
+    if (process.argv[2] === 'deploy' && !hasServerlessConfigFile(process.cwd())) {
+      return true;
+    }
+
+    if (!componentConfig && !instanceConfig && process.argv.length === 2) {
+      return true;
+    }
+
+    if (runningTemplate(process.cwd())) {
```
medikoo, line 59: I've looked closer, and we already have `runningTemplate` detection here.

It just additionally confirms that component command is used: https://github.com/serverless/components/blob/a8633ca7ed88a7919efe777860d5c10861ef4bd9/src/legacy.js#L32

I think best would be to simply update thise method, so also `dev` is recognized as component command

medikoo (changes requested):
@zongUMR sorry for second thought, but I've noticed that template check is already made. Please see my comment

(note this check can be expensive, so it's better to not double that)

src/legacy.js
zongUMR: But if we have more commands later, we must add them in `nestedTemplateKeywords`. For Chinese users, if they are in template, just need to load components I think.

zongUMR: And I also have a question, why we can not just use `runningTemplate`, and use `nestedTemplateKeywords` to limit the methods?

medikoo: As this check expensive we decided to imply that check only if used command is recognized as component command.
And same problem applies if user is in China, I believe same rule should be followed in case of Chines version.

Not this check does recursive deep dir read, which in directory with big structure may turn to be time-taking.

medikoo (changes requested):
As this check  is expensive we decided to imply that check only if used command is recognized as component command.
And as same problem applies if user is in China, same rule should be followed

Note this check does recursive deep dir read, which in directory with big structure may turn to be time-taking.

src/legacy.js
medikoo: > and use nestedTemplateKeywords to limit the methods?

What do you mean by that?

zongUMR: I mean we use `nestedTemplateKeywords` to limit only those methods in **template project** will load components, right?

medikoo: `nestedTemplateKeywords ` is the list of component commands that can be used for template projects.

So first we validate if given command applies to template project, and only having that confirmed we check whether we are in template project (and that check in some edge scenarios may lock the process for a while)

zongUMR: So, your suggestion is that we do not change the logic here, only update `nestedTemplateKeywords` with the new commands we want to run in template?

medikoo: > So, your suggestion is that we do not change the logic here, only update nestedTemplateKeywords with the new commands we want to run in template?

yes, exactly

zongUMR: In this respect. I agree with this view. @ole3021 How do you think about it?

ole3021:
@medikoo  Thanks for the update about the logic, and the story behind this is 

* currently we are adding more commands into Serverless Tencent, as `invoke`, `invoke local` and `logs` and we are build the feature to support those commands from the template folder.  (as the template folder is a set of components).
    * Without the support of those commands from template folder, when user do multi components development, user needs to go to each compoents folder to using the commands, which is not a good experience for developers, as it distract users from their own code and logic. 
* Users also complained about the behave different in template folder when execute some Serverless Tencent commands, as except all the commands are handled by the Serverless Traditional. (and user dont have a good understanding about components and template.)
    * also we have a command not support chinese warning message for users when they execute some non exist commands, but this never works in template folder. 

* Also based on the discussion we aligned earlier, we will keep the Serverless Traditional for users in China as default and for the traditional projects by default. (without set the vendor in ENV)

I think compares to add those new commands in the support list, its better to handle the template folder with Serverless Tencent instead. 

If it is only the performance issue block us to do so, can we change the improve the logic about checking if a folder is a Template Folder? and here are my suggestions.

## solution A : Improve the template check

> A template folder should be a folder have `serverlss.yml` with only common fields (no component, no inputs, no functions.) And at least a sub components folder in it (have serverless.yml and include component field.)

so can we check as the follow process
1. if current folder have `serverless.yml` with out fields (components, inputs, function)
   * [yes] go step 2
   * [no] false
2. get the nest folders in current folder. (move probably system folder to the last , eg .git, node_modules etc)
3. check nest folder one by one, if have serverless.yml and if there have component exist
   * [yes] true 
   * [no] go 3
   * [chkeck all the nest folders] go 4
4. false.

or even simpler
Just check if current folder have `serverless.yml` (true) and no fields in it (components, inputs, functions, etc), then consider it as a template folder.

## Solution B : Replace check a folder is template , check a folder is traditional.

> as China users use Serverless Traditional is not much, instead so spending time verify if a folder is template, we can check if a folder is Traditional instead. 

Here are the logic in detail. 

isChinaUser: true (User from China)
* If current folder has no serverless.yml.yaml.json.js etc, take it as the empty folder, and handls by Serverless Traditional by default. 
* If a folder has serverless.yml, and there have functions fields in it, handls it by Serverless Traditional as it is a traditional folder. 
* for the left are handles by Serverless Tencent. 

Please have a check , and im love to have future talk with you when you got tim.

medikoo:
@ole3021 thanks for explanations.

> I think compares to add those new commands in the support list, its better to handle the template folder with Serverless Tencent instead.

Yes, it's better to invoke _runningTemplate_ check only if we confirmed that we're in context of command that's supported by Components template project.

>If it is only the performance issue block us to do so, can we change the improve the logic about checking if a folder is a Template Folder?

Current implementation of this check I believe is as optimal as it can be. You can investigate it here:https://github.com/serverless/components/blob/a8633ca7ed88a7919efe777860d5c10861ef4bd9/src/cli/utils.js#L548-L578

I've check the options you list, but they seem to offer quite different resolution logic, that will actually change the detection and not optimise it.

Current rules that establish what is template project are as follows (all must be met):
- There's no service configuration (`serverless.yml` or similar) at root folder
- One of the child folders (of which name doesn't start with `.`) should have component configuration (so `serverless.yml` file or similar, which exports `component` property)
- There must be no child folder which contains Framework service configuration (it's to prevent situation where we mark as template, something as _Projects_ folder which may contain many different component projects and regular framework projects as well)

If you have suggestions how to better implement solutions that confirms on that rules, it's definitely welcome, but otherwise I think that those rules should not be changed

ole3021:
@medikoo  thx for update the logic, and i do miss check the code before propose the solution. Here are the logic table for use cases (folder) and conditions below. The priority order is from left to right.

|                        | Non-SLS     | Components | Traditional | Template   | Other       |
| ---------------------- | ----------- | ---------- | ----------- | ---------- | ----------- |
| isChinaUser            | -           | true       | -           | true       | -           |
| root serverless.yml    | false       | true       | true        | true       | true        |
| field: components      | -           | true       | false       | false      | -           |
| field: services        | -           | -          | true        | false      | -           |
| > deployable           | -           | true       | true        | false      | -           |
|                        |             |            |             |            |             |
| nest serverless.yml    | -           | -          | -           | least 1    |             |
| nest field: components | -           |            |             | all true   | -           |
| nest field: services   | -           |            |             | all false  | -           |
|                        |             |            |             |            |             |
| use CLI                | traditional | components | traditional | components | traditional |
|                        |             |            |             |            |             |

true: means need to exist .
false: meens need to not exist.
-: means no need to consider

Base on the table above, i think the issue we are facing here is to tell template correctly from Other folders include serverless.

Then the logic `runningTemplate ` above only purpose is to tell template from other folders. 
if we can agree on this. i think we can do following changes to easy solve the issue with out heavy calculation. 

* if a serverless folder (include serverless) is not `deployable`.
    * Do not do nest folder calculation. 
    * Use components directly for China users, and keep traditional for non-China users.

So we can avoid the nest folder check, and for the template users case, it will work the same, but for the other user cases, changes will be 
* [isChinaUser] will show components error message and warnings.
* [other] will show traditional error message. 

Thats what we want to achieve.

medikoo:
@ole3021 Thanks for explanations.

If I understand above table correctly. The difference between _template_  project of AWS and of Tencent components. Is that AWS template project does not host root `serverless.yml`. While Tencent template project hosts it. Is that the case?

If so, what's the content of root `serverless.yml` in such case?

_Note: Otherwise I do not see any differences, and so far logic should work well (aside of fact that listed commands need to reflect also all commands that are supported in Tencent template projects_

zongUMR:
@ole3021 @medikoo I will only add new commands in `legacy.js` for this PR to enable it available. We can still discuss this topic in this thread, and we can improve this logic once we have a result

timqian (approved):
The current implementation looks good to me

PR #998, opened 2021-07-21
Detect invoking dir for templates

timqian:
When users execute `sls invoke` or `sls invoke local` in a template folder, try to detect `scf`/`multi-scf` instance directory.

### Related ticket
https://app.asana.com/0/1200011502754281/1200578828969029

### DIFF
--- a/src/cli/commands-cn/utils.js
+++ b/src/cli/commands-cn/utils.js
@@ -208,6 +208,66 @@ const loadInstanceCredentials = () => {
   return credentials;
 };
 
+const getDirForInvokeCommand = async (root, functionAlias) => {
+  let instanceDir = root;
+  const instances = [];
+
+  const directories = fs
+    .readdirSync(root)
+    .filter((f) => fs.statSync(path.join(root, f)).isDirectory());
+
+  for (const directory of directories) {
+    const directoryPath = path.join(root, directory);
+
+    const instanceYml = loadInstanceConfig(directoryPath);
+
+    if (instanceYml) {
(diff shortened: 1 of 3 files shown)


src/cli/commands-cn/utils.js
```diff
@@ -210,6 +210,66 @@ const loadInstanceCredentials = () => {
...
+        instanceInfo.functions = Object.keys(instanceYaml.inputs.functions);
+      }
+      instances.push(instanceInfo);
+    }
+  }
+
+  const multiScfInstances = instances.filter((instance) => instance.component === 'multi-scf');
+  const scfInstances = instances.filter((instance) => instance.component === 'scf');
+  if (scfInstances.length + multiScfInstances.length === 0) {
+    throw new Error('没有找到可执行的函数目录，请使用 --target 指定或检查后再试');
```
zongUMR, line 248: Should we use warnings instead of erro here and elsewhere?

timqian: The error thrown here will be caught here: https://github.com/serverless/components/pull/998/files#diff-2a4f4f9def1700b7cc7cf1d95bf2250063ea626cf3bc4a69101cc51cefbe796fR59 and users will see warnings.

zongUMR (approved):

zongUMR:
Does `logs` command need this feature?

timqian:
@zongUMR This PR is only for `invoke` and `invoke local` which are only related to SCF-related components. We can discuss later if we want to add similar behavior to commands like `logs`, `dev`, and other general commands.

PR #1011, opened 2021-08-18
fix `=` sign  for inputs value

zongUMR:
1. Add `credentials` command in the template project
2. Fix `=` sign in inputs value


timqian (approved):
Looks good to me

PR #1031, opened 2021-11-16
feat: upgrade generateYMLForNodeProject on http component

zongUMR:
## What has been implemented?

Closes https://app.asana.com/0/1200011502754281/1201300336822086/f

## Steps to verify

- `npm i @serverless/components@canary -g`
- `components init eggjs-starter(express-starter, koa-starter, nextjs-starter, nuxtjs-starter) --name example`
- `cd example`
- ` rm serverless.yml`
- `components deploy`: CLI will automatically generate a new `serverless.yml` for this project via the `http` component and deploy successfully

### DIFF
--- a/src/cli/commands-cn/utils.js
+++ b/src/cli/commands-cn/utils.js
@@ -387,84 +387,100 @@ const saveYaml = async (yamlPath, yamlObj) => {
 };
 
 const generateYMLForNodejsProject = async (cli) => {
-  const getExpressYML = (entryFile) => `component: express
+  const getExpressYML = () => `
+component: http
 name: expressDemo
-app: appDemo
 
-inputs:${entryFile ? `\n  entryFile: ${entryFile}` : ''}
-  src: ./
-  region: ap-guangzhou
-  runtime: Nodejs10.15
-  apigatewayConf:
+inputs:
+  src:
+    src: ./
+    exclude:
+      - .env
+  faas:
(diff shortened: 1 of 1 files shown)


zongUMR:
@timqian Please correct me if I am wrong, I removed the last line `throw Error` in the `generate` function, because I think it's impossible to reach out to that error because of the previous checks, and we can finally return the `egg` YML in the end.

src/cli/commands-cn/utils.js
```diff
@@ -536,11 +547,8 @@ inputs:
     return getNuxtYML();
   }
 
-  if (ymlType === 'egg') {
-    return getEggYML();
-  }
-
-  throw new Error('当前目录未检测到 Serverless 配置文件');
+  // finally return egg's yml
+  return getEggYML();
```
timqian, line 551: Why are we returning YML for egg component here for all cases?

zongUMR: We don't return it for all, only for `egg`, due to we have a checker for known packages? https://github.com/serverless/components/pull/1031/files#diff-0e74e48ad16540e1bf681738cd97a41ea2fa6cbdfd5d987b62208fc1a7c829f8R499

timqian: Yeah, your are right👍

timqian (approved):

